PR #2, opened 2020-09-16
Adds roles to users and enables editing of positions

yhedholm:
Change 1:
The old roles system didn't work: it didn't store properly and it didn't have enough roles. Importantly, there was no Dancer role and AA has asked for filtering on Dancers when specifying casts. On top of this, what the frontend UI called Permissions, the backend called Roles. Finally, phoneNumbers were not stored. As they are important for messaging, it made sense to add these when we changed the backed to support the new roles.
Change 2:
The Positions could not be edited, and they could not be deleted (and added back again -- a "poor person's editing" once casts or performances had bee created around them. Spelling errors need to be correctable so we needed to fix this.

Because of the amount of boiler plate code in Spring Boot every database change forces code changes in multiple files. This looks like a huge change but it really isn't. The same changes, adding the phonNumber field and the new role fields, are repeated over the jasonobjects files, model files, and frequently the services files, the controllers, and test files.

On the backend these changes were implemented:
1) phoneNumber was added to the User table.
2) The old Roles class was renamed Permissions and the Admin item was removed from Permissions.
3) A new Roles class was created that contains the roles isAdmin, isCoreographer, isDancer, and isOther.
4) The role field was removed from the User table.
5) I also changed the order of some of the fields in the User table which causes some extra diffs.

On the frontend, the implementation of these changes involved removing two privilege_class_api.service files in the api directory. They were no longer called. I feel all other frontend changes are straight-forward. The diff is a little larger because I reformatted some lines, mostly because lines were too wide, but sometimes because I thought I made them easier to read.

When it comes to Change 2, it was front end only and affected mostly the piece_api.services.ts file in the api directory (mostly reformatting) and the files in the piece_editor.component files in the piece directory.


backend/src/main/java/com/google/rolecall/services/UserServices.java
```diff
@@ -96,21 +99,25 @@ public User editUser(UserInfo newUser) throws EntityNotFoundException, InvalidPa
...
         .setManagePieces(newUser.managePieces())
         .setManageRoles(newUser.manageRoles())
-        .setManageRules(newUser.manageRules());
-
+        .setManageRules(newUser.manageRules())
+        .setEmergencyContactName(newUser.emergencyContactName())
+        .setEmergencyContactNumber(newUser.emergencyContactNumber())
+        .setComments(newUser.comments())
+        .setIsActive(newUser.isActive())
+        ;
```
zephyr-l, line 120:
```suggestion
        .setIsActive(newUser.isActive());
```

backend/src/test/java/com/google/rolecall/restcontrollers/UserManagementUnitTests.java
```diff
@@ -61,22 +64,26 @@ public void init() {
...
+      .setManagePerformances(managePerformances)
+      .setManageCasts(manageCasts)
+      .setManagePieces(managePieces)
+      .setManageRoles(manageRoles)
+      .setManageRules(manageRules)
+      .setEmergencyContactName(emergencyContactName)
+      .setEmergencyContactNumber(emergencyContactNumber)
+      .setComments(comments)
+      .setIsActive(isActive)
+      ;
```
zephyr-l, line 86:
```suggestion
      .setIsActive(isActive);
```

backend/src/main/java/com/google/rolecall/jsonobjects/UserInfo.java
```diff
@@ -157,6 +170,18 @@ public static Builder newBuilder() {
     @JsonProperty("manageRules")
     public abstract Builder setManageRules(Boolean manageRules);
 
+    @JsonProperty("emergencyContcatName")
+    public abstract Builder setEmergencyContactName(String emergencyContcatName);
+
+    @JsonProperty("emergencyContactName")
+    public abstract Builder setEmergencyContactNumber(String emergencyContactNumber);
```
zephyr-l, line 177: Just FYI, I think in the frontend we do not have field to display these 2 fields.

zephyr-l (approved):
👍

backend/src/main/java/com/google/rolecall/jsonobjects/UserInfo.java
zephyr-l: As I am not very familiar with Typescript, I only reviewed the backend code and they look good to me.

backend/src/main/java/com/google/rolecall/services/UserServices.java
yhedholm: Zephyr, the reason for placing the semicolon on a separate line is that the semicolon is the one asymmetry that makes it harder to add another field initialization after setIsActive(newUser.isActive());. If you try to, and forget to remove the semicolon, you get a syntax error. By moving the semicolon to a separate line, you can add initializers or move the initializers around freely without worrying where the semicolon is. For the same reason, I try to adda comma after every field definition in typescript, even after the last one that is optional.

backend/src/test/java/com/google/rolecall/restcontrollers/UserManagementUnitTests.java
yhedholm: See my comment above.

backend/src/main/java/com/google/rolecall/jsonobjects/UserInfo.java
yhedholm: Zephyr, yes the emergency fields are not implemented in the user interface yet. The are other fields that are implemented on the backend but not implemented on the front end yet.

frontend/rolecall/src/app/piece/piece_editor.component.scss
```diff
@@ -54,16 +50,33 @@
...
+.add-button-width-container-right {
   display: flex;
   flex-direction: row;
   justify-content: flex-end;
-  min-height: 10vh;
+  min-height: 8vh;
+  width: 99%;
+}
+.add-button-height-container-left {
+  //background: red;
```
mikzat, line 69: remove commented out code

frontend/rolecall/src/app/piece/piece_editor.component.ts
```diff
@@ -20,12 +27,12 @@ type WorkingPiece = Piece & {
...
 export class PieceEditor implements OnInit {
 
-  dragAndDropData: { type: "adding" | "added", index: number, value: Position }[] = [];
+  dragAndDropData: DraggablePosition[] = [];
   currentSelectedPiece: WorkingPiece;
   renderingPieces: WorkingPiece[];
   urlPointingUUID: string;
 
-  privilegeClasses: string[] = [];
+  //privilegeClasses: string[] = [];
```
mikzat, line 35: remove

frontend/rolecall/src/app/user/user-editor.component.ts
```diff
@@ -18,27 +17,24 @@ import { User, UserApi } from '../api/user_api.service';
 export class UserEditor implements OnInit {
 
   permissionsSet: EventEmitter<string[]> = new EventEmitter();
-  privilegeClassSet: EventEmitter<string[]> = new EventEmitter();
+  rolesSet: EventEmitter<string[]> = new EventEmitter();
 
   currentSelectedUser: User;
   renderingUsers: User[];
+  su: User[]; // Sorted Users
```
mikzat, line 24: Please call this "sortedUsers" to keep var names readable.

frontend/rolecall/src/app/user/user-editor.component.ts
```diff
@@ -291,42 +296,44 @@ export class UserEditor implements OnInit {
...
-    else if (key == "Privilege Classes") {
-      if (this.workingUser.has_privilege_classes) {
-        let largerArr = val.length > this.workingUser.has_privilege_classes.length ? val : this.workingUser.has_privilege_classes;
-        let smallerArr = val.length <= this.workingUser.has_privilege_classes.length ? val : this.workingUser.has_privilege_classes;
-        for (let i = 0; i < largerArr.length; i++) {
-          if (largerArr[i] != smallerArr[i]) {
-            this.userSaved = false;
-            break;
+    } else if (info.type == "roles") {
+      let val2 = this.workingUser.has_roles;
```
mikzat, line 318: 1. Let's avoid variable names like "val" and "val2" where possible and name it something more semantic
2. Let's make this a const since it is not reassigned
3. `has_roles` makes me expect a boolean true/false, but it looks like it contains a list? Seems surprising.

frontend/rolecall/src/app/user/user-editor.component.ts
```diff
@@ -291,42 +296,44 @@ export class UserEditor implements OnInit {
...
+      for (let entry of Object.entries(val2)) {
```
mikzat, line 319: Rather than `entry` let's use a var name that describes what the entry is.

mikzat (changes requested):
This is awesome Yorgen, thanks for all the changes/cleanup! As I mentioned in chat, it would be great in the future to separate cleanup/moving things as a separate PR, and logic/functional changes in a PR by themselves. It would really help speed up the review process!

backend/src/main/java/com/google/rolecall/jsonobjects/UserInfo.java
```diff
@@ -114,31 +128,30 @@ public static Builder newBuilder() {
...
-    public abstract Builder setRole(String role);
+    @JsonProperty("isDancer")
+    public abstract Builder setIsDancer(Boolean isDancer);
 
-    @JsonProperty("isActive")
-    public abstract Builder setIsActive(Boolean isActive);
+    @JsonProperty("isOther")
+    public abstract Builder setIsOther(Boolean isOther);
 
+    // Permissions
```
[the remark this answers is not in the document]
yhedholm: Yes, it would be a significant change. Also, given how "under designed" this project is, having Roles and Permissions un-nested gives us more flexibility to get the right permissions "is spite" of whatever roles are present.

frontend/rolecall/src/api_types.ts
```diff
@@ -9,27 +9,31 @@ export enum PerformanceStatus {
...
-    canLogin: boolean,
-    isAdmin: boolean,
-    notifications: boolean,
-    managePerformances: boolean,
-    manageCasts: boolean,
-    managePieces: boolean,
-    manageRoles: boolean,
-    manageRules: boolean
+    canLogin: boolean;
+    notifications: boolean;
```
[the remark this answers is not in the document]
yhedholm: I assume it marks that someone can receive notifications. The notification part has not been implemented yet.

backend/src/main/java/com/google/rolecall/services/UserServices.java
zephyr-l: I referred to the Google Java Style Guide and did not see this style breaking rules. Although it is not very common but I think I will take it.

Let's all use the Google Java Formatter https://github.com/google/google-java-format#using-the-formatter to keep our format consistent.

backend/src/main/java/com/google/rolecall/jsonobjects/UserInfo.java
zephyr-l: Thanks for confirming that.

frontend/rolecall/src/app/user/user-editor.component.ts
yhedholm: Agree 100%. However, fixing it is not trivial.  As I mentioned above, I decided to copy the functionality of permissions, that already worked fine, when I fixed the broke role system. The code at line 317 is identical to the code at 311 except 311 pertains to Permissions and 317 pertains to Roles.  All I did when I added this code was to change Permissions to Roles and to verify that all variables I used were local. I was right, the code works, and because it is logically identical, it should make the program more difficult to understand. On the down-side, I also inherited whatever bad programming practices were used before.

frontend/rolecall/src/app/user/user-editor.component.ts
yhedholm: Yes. But see comment above.

backend/src/test/java/com/google/rolecall/restcontrollers/UserManagementUnitTests.java
yhedholm: Given that Mikhail also didn't like the trailing ; I am revering that change.

backend/src/main/java/com/google/rolecall/Constants.java
```diff
@@ -32,16 +32,25 @@
 
   public static class Roles {
```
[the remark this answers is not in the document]
yhedholm: Naming-wise I mirrored exactly what was done with Permissions (including the names for all supporting functions). If this is a trivial change, we should consider it. Otherwise it should be twice the effort because we should do Permissions as well.

frontend/rolecall/src/api_types.ts
```diff
@@ -9,27 +9,31 @@ export enum PerformanceStatus {
...
   export type CastUUID = string;
   export type PerformanceUUID = string;
   export type UnavailabilityUUID = number;
   export type SuccessIndicator = {
-    successful: boolean,
-    error?: string
+    successful: boolean;
+    error?: string;
+  };
+  export type RoleSet = {
```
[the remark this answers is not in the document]
yhedholm: I changed RoleSet to Roles and PermissionSet to Permissions.

frontend/rolecall/src/api_types.ts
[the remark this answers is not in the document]
yhedholm: I changed notifications to canReceiveNotifications.

backend/src/main/java/com/google/rolecall/services/UserServices.java
mikzat: Thanks for the link, I was going to propose using [EditorConfig](https://editorconfig.org/) but maybe these are complementary. I'll take a look at the plugin you suggested.

backend/src/main/java/com/google/rolecall/Constants.java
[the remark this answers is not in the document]
mikzat: Ok, let's leave it as-is for now.

frontend/rolecall/src/app/user/user-editor.component.ts
mikzat: Acknowledging that we are inheriting a lot of this stuff, and we don't have to fix everything all at once.

frontend/rolecall/src/app/api/user_api.service.ts
```diff
@@ -10,67 +10,51 @@ import { LoggingService } from '../services/logging.service';
...
+  has_roles: APITypes.RoleSet;
+  has_permissions: APITypes.PermissionSet;
+  knows_positions: APITypes.Position[];
+  first_name: string | undefined;
+  last_name: string | undefined;
+  date_joined: number | undefined;
   contact_info: {
-    phone_number: string,
-    email: string,
+    phone_number: string | undefined;
```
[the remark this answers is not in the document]
yhedholm: The email is a required field. The phone number is not required now, but it could change going forward.
However, this data structure is now supporting user input, so it has to allow for undefined values. There is no field-is-defined check on the front end. That checking is done on the backend.

frontend/rolecall/src/app/api/user_api.service.ts
```diff
@@ -10,67 +10,51 @@ import { LoggingService } from '../services/logging.service';
...
-  admin?: boolean,
-  notifications?: boolean,
-  managePerformances?: boolean,
-  manageCasts?: boolean,
-  managePieces?: boolean,
-  manageRoles?: boolean,
-  manageRules?: boolean,
-  isActive?: boolean
+type RawAllUsersResponse = {
+  data: rawUser[],
```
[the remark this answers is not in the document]
yhedholm: Probably. I don't know how much work would be required.
The change I did was to replace a separate definition of raw user
data: {
  id: number,
  firstName: string,
  ...
}
with the new standard definition of rawUser (raw marks data we get from the backend). The old code was on line 31 in the old file and in the compare. The old team loved to call data received from the backend data. Going forward I'd like to change it, but it is probably not a priority.

frontend/rolecall/src/app/cast/cast-drag-and-drop.component.ts
```diff
@@ -253,6 +253,7 @@ export class CastDragAndDrop implements OnInit {
           return castPositionOrerA < castPositionOrerB ? -1 : 1
         });
     let filledPoses = this.cast.filled_positions;
+    // XXXXX
```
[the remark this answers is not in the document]
yhedholm: Done.

frontend/rolecall/src/app/piece/piece_editor.component.scss
yhedholm: Done.

frontend/rolecall/src/app/piece/piece_editor.component.ts
yhedholm: Done.

frontend/rolecall/src/app/user/user-editor.component.ts
yhedholm: When I went to fix this, I realized that the variable wasn't used. (I was only used for a debug printout earlier.) This inspired me to try to lint the project.
Tons of problems: missing semicolons, let that should be const, etc. etc.
I am tempted to do a separate PR for fixes to the lint errors.

mikzat: Yeah we'll fix the lints and nits as we go :)

mikzat (approved):

frontend/rolecall/src/app/user/user-editor.component.ts
yhedholm: in my earlier comment, I meant "... it should NOT make the program more difficult to understand."

PR #4, opened 2020-09-21
Adding images

mikzat:
Two separate commits here for easier code review:
1. Cleanup of relevant files
2. Adding image support based on first name and last name

chrome-capture (3)

### DIFF
--- a/frontend/rolecall/src/app/cast/cast-drag-and-drop.component.spec.ts
+++ b/frontend/rolecall/src/app/cast/cast-drag-and-drop.component.spec.ts
@@ -1,6 +1,5 @@
-import { async, ComponentFixture, TestBed } from '@angular/core/testing';
-import { CastDragAndDrop } from './cast-drag-and-drop.component';
-
+import {async, ComponentFixture, TestBed} from '@angular/core/testing';
+import {CastDragAndDrop} from './cast-drag-and-drop.component';
 
 describe('CastDragAndDropComponent', () => {
   let component: CastDragAndDrop;
@@ -8,8 +7,8 @@ describe('CastDragAndDropComponent', () => {
 
   beforeEach(async(() => {
     TestBed.configureTestingModule({
-      declarations: [CastDragAndDrop]
-    })
+        declarations: [CastDragAndDrop]
+      })
       .compileComponents();
   }));
 
--- a/.gitignore
+++ b/.gitignore
@@ -5,4 +5,4 @@ backend/.vscode/*
 backend/.settings/*
 backend/.classpath
 backend/.factorypath
-
+.idea/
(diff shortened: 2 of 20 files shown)


yhedholm (commented):
Mihail, I went through this carefully as I was trying to learn the rules you use as I will try to use them as well. It seems as you prefer
1) deep rather than wide statements
2) No space around import {} brackets
3) single quotes '
4) === vs ==
5) const whenever possible

I will try to work the way you worked with making cleanup changes first and then the code changes. However, for the current changes made under duress and pressure, I will delay the cleanup.

frontend/rolecall/src/app/cast/cast-drag-and-drop.component.spec.ts
```diff
@@ -1,15 +1,14 @@
...
 
 describe('CastDragAndDropComponent', () => {
   let component: CastDragAndDrop;
   let fixture: ComponentFixture<CastDragAndDrop>;
 
   beforeEach(async(() => {
     TestBed.configureTestingModule({
-      declarations: [CastDragAndDrop]
-    })
+        declarations: [CastDragAndDrop]
```
yhedholm, line 10: I am trying to understand the logic for this indentation change. Is 'declaration' indented once for the '(' and once for the '{'?

yhedholm (approved):

frontend/rolecall/src/app/cast/cast-drag-and-drop.component.spec.ts
mikzat: That's an interesting case, I had to double-check it myself. Looks like it's not from the parenthesis, but because it is followed by the chained `.compileComponents()` call, which caused a 4-space indent and this line indented to line it up. Without a chained call it ends up being just the regular 2-space indent.

mikzat:
> Mihail, I went through this carefully as I was trying to learn the rules you use as I will try to use them as well. It seems as you prefer
> 
> 1. deep rather than wide statements
> 2. No space around import {} brackets
> 3. single quotes '
> 4. === vs ==
> 5. const whenever possible
> 
> I will try to work the way you worked with making cleanup changes first and then the code changes. However, for the current changes made under duress and pressure, I will delay the cleanup.

I basically tried to export/import our internal conventions into my IDE and ended up with the cleanup. It's not 100% matching our internal style guides, but pretty close. HTML is not formatted quite the way I want (needs a 4-space line indent rather than 2 on attributes/properties), but I don't want to tinker too much with it right now.

Yes those are few of the rules, #2-5 match up with our internal conventions and avoiding long lines is a good practice for better readability. A long line forces reader to read almost completely from left to right to find what you're looking for, but by limiting to only a couple of "concepts" on a single line allows to more quickly "visually parse" the logic, plus it makes it easier to diff changes in PR across multiple lines v.s. one long line, and easier to see history of who changed which line last.

mikzat:
Scrapping this PR, having too much trouble resolving the merge conflicts and it's faster to just do a clean change from scratch.

PR #5, opened 2020-09-23
Adding cast images

mikzat:
Adding image support based on first name and last name.

image


yhedholm (approved):
That was an easy review!

PR #6, opened 2020-09-24
Minor cleanup on root level of frontend code while I explore how the app is built

mikzat:
There should be no functional changes from this. If you suspect there might be, please point that out in review.


yhedholm (approved):
All changes look fine to me. There will probably be some incompatibilities but they should not be hard to fix.

PR #7, opened 2020-09-25
Almost only backend changes for revelation support

yhedholm:
These are the backend changes made to enable support for 'revelations'. The only frontend file included is the file that reads the performances, where a data field had to be added.

In ApplicationLoader.java I added sample data (the Alvin Ailey users are added whenever the database is empty. In order to give them a starting data, I had to add some date parsing code.)

When I added the positionOrder field in /jsonobjects/PerformancePositionInfo.java, I got a syntax error in /jsonobjects/CastMemberInfo.java that I couldn't figure out how to get rid of. I never liked the different structure of the old CastMemberInfo file anyway, so I rewrote it using a standard builder pattern. Because of this, I also had to change the builder call in models/CastMember.java.

The other changes are logic changes necessary store and read revelations.

### DIFF
--- a/backend/src/main/java/com/google/rolecall/ApplicationLoader.java
+++ b/backend/src/main/java/com/google/rolecall/ApplicationLoader.java
@@ -3,9 +3,15 @@ package com.google.rolecall;
 import com.google.rolecall.models.User;
 import com.google.rolecall.repos.UserRepository;
 import com.google.rolecall.restcontrollers.exceptionhandling.RequestExceptions.InvalidParameterException;
+
+import java.text.ParseException;
+import java.text.SimpleDateFormat;
+import java.util.Calendar;
+import java.util.Locale;
 import java.util.Optional;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.ApplicationArguments;
 import org.springframework.boot.ApplicationRunner;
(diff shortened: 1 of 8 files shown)


backend/src/main/java/com/google/rolecall/ApplicationLoader.java
```diff
@@ -61,6 +67,86 @@ private void createAdmin() {
...
+      userRepo.save(user);
+    } catch(InvalidParameterException e) {
+      return DataCreateError.OtherError;
+    } catch(ParseException e) {
+      return DataCreateError.OtherError;
+    }
+    return DataCreateError.OK;
+  }
+
+  private DataCreateError createTestData() {
```
zephyr-l, line 105: I do not think the createTestData method here fits the ApplicationLoader class. There are 2 reasons:

1. All tests related data should in javatests folder. We should not use this in production environment. We have the name but other information including the email and dates are incorrect. The email is an immutable field in the frontend.
2. This will be called every time we have a database refresh and there is no way to opt out and create a clean database.

Some suggestion about the approach:
1. Make the data seeding as part of the javatests and totally remove it from the java folder. (recommended)
2. If we have to seed data, then at least we should mark it with environments and only do it in non-prod.
3. And, we should initiate some controls about when to seed the data, such as creating a flag or some environment variables.

backend/src/main/java/com/google/rolecall/jsonobjects/CastMemberInfo.java
```diff
@@ -27,14 +28,42 @@
...
+  @JsonProperty("delete")
   public abstract Boolean delete();
 
-  @JsonCreator
-  public static CastMemberInfo create(@Nullable @JsonProperty("id") Integer id, 
-      @Nullable @JsonProperty("userId") Integer userId,
-      @Nullable @JsonProperty("subCastId") Integer subCastId,
-      @Nullable @JsonProperty("order") Integer order,
-      @Nullable @JsonProperty("delete") Boolean delete) {
-    return new AutoValue_CastMemberInfo(id, userId, subCastId, order, delete);
```
zephyr-l, line 38: We can have a create method call the builder and pass all the values to keep it backward compatible, but I am fine if we make these changes.

backend/src/main/java/com/google/rolecall/models/CastMember.java
```diff
@@ -45,7 +45,13 @@ public SubCast getSubCast() {
...
 
   public CastMemberInfo toCastMemberInfo() {
-    return CastMemberInfo.create(id, user.getId(), cast.getId(), order, null);
+    return CastMemberInfo.newBuilder()
+    .setId(id)
+    .setUserId(user.getId())
+    .setSubCastId(cast.getId())
+    .setOrder(order)
+    .setDelete(null)
+    .build();
```
zephyr-l, line 54: There're some formatting issues. 

I strongly encourage use https://github.com/google/google-java-format to format the code.

backend/src/main/java/com/google/rolecall/services/PerformanceServices.java
```diff
@@ -202,7 +202,7 @@ private PerformanceSection addPerformanceCastMember(PerformanceSection performan
     User user = userService.getUser(memberInfo.userId());
 
     PerformanceCastMember member = PerformanceCastMember.newBuilder()
-        .setOrder(position.getOrder())
+        .setOrder(memberInfo.order())
```
zephyr-l, line 205: Is this the cast position? (1st cast, 2nd cast, etc.)

backend/src/main/java/com/google/rolecall/models/PerformanceSection.java
```diff
@@ -138,8 +141,9 @@ public PerformanceSectionInfo toPerformanceSectionInfo() {
   }
 
   private List<PerformancePositionInfo> toAllPerformancePositionInfo(
-      Hashtable<Integer, Hashtable<Integer, List<PerformanceCastMemberInfo>>> positions) {
-    List<PerformancePositionInfo> positionInfos = new ArrayList<>();
+      Hashtable<Integer, Hashtable<Integer, List<PerformanceCastMemberInfo>>> positions,
+      Hashtable<Integer, Integer> positionOrders) {
+  List<PerformancePositionInfo> positionInfos = new ArrayList<>();
```
zephyr-l, line 146: Please check the indention.

backend/src/main/java/com/google/rolecall/ApplicationLoader.java
```diff
@@ -61,6 +67,86 @@ private void createAdmin() {
...
+    logger.log(Level.WARNING, "Creating test data.");    
+    Boolean finished = false;
+    while (!finished) {
```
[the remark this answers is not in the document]
yhedholm: You are right. I shouldn't have captured the error in the 'lower' function and instead handled it in the 'higher' function with the added benefit that the higher level function would be cleaner.

On the other hand, the only way for me to test any code changes here is to wipe the database (to your point below) repeatedly, and right now I have created good demo data.  And, I will try to explain below, this code will have a limited life span (I plan to remove it as soon as Alvin Ailey is up and running).

Rather than fixing the code, I vote that we leave it in for now (it works after all) and remove it in a week or two.

backend/src/main/java/com/google/rolecall/ApplicationLoader.java
yhedholm: Zephyr, you are right about everything you say. Yet, I would prefer to leave the code in for the next few weeks.

We are writing a system for one client and initially maybe two or three users (who will plan the tours of a little over 30 dancers). For now, this is a tiny, one-off system. We want to make the users' startup experience as good as possible. For a non trained typist, entering 31 names with sometimes non standard spelling can be challenging. (I made several mistakes myself when I entered the same data.) Especially given that the names have to be spelled correctly -- and the middle names and suffixes have to be dealt with correctly -- for the picture lookup to work. I believe having the dancers pre loaded will be viewed as a big plus.

Your other point is true too. We need to get the correct email addresses into the system. (Probably not for login purposes, but for notifications.) And given that we don't have them now, and the email field is not editable, we may have to insert the proper email addresses using SQL later.

Anyway, even though this code probably breaks a lot rules for production code, I lobby that we leave it in. The Alvin Ailey project is a very non standard project.

backend/src/main/java/com/google/rolecall/jsonobjects/CastMemberInfo.java
yhedholm: I agree. I spoke to Jared about this just before he left and he told me he was just experimenting, but in retrospect he wished he had used the builder patter here too.

backend/src/main/java/com/google/rolecall/models/CastMember.java
yhedholm: Fixed. I installed the google formatter. Thanks for the recommendation!

backend/src/main/java/com/google/rolecall/services/PerformanceServices.java
yhedholm: No, this is the dancer # field when you have many dancers in the same position. I thought this field was unused because it was always 0 with the test data I had. (So I tried to reuse it for position order.) This change restores the original code before my position order change. (I now get the position order from the code change below.)

backend/src/main/java/com/google/rolecall/models/PerformanceSection.java
yhedholm: Fixed.

backend/src/main/java/com/google/rolecall/jsonobjects/CastMemberInfo.java
zephyr-l: Sounds good. Thanks for letting me know.

backend/src/main/java/com/google/rolecall/ApplicationLoader.java
zephyr-l: I realized the pain point about typing it over and over. How about we do this:

1. Pack this seeding procedure into a dedicated method.
2. Add a TODO saying this is for experimental only and needs to be removed.

backend/src/main/java/com/google/rolecall/ApplicationLoader.java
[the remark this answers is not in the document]
zephyr-l: I think the fix should be relatively simple.

In createOneUser, let's throw exception if the creation failed.

In createTestData, let use

```
try {
  createOneUser("Robert", "Battle", "rb@gmail.com","1/1/2020"));
  createOneUser("Jeroboam", "Bozeman", "bb@gmail.com","1/1/2020"))
  ...
} catch (SomeException e) {
  // handle the exception such as logging the error
}
```

I am just not in favor of the while loop. Seems it is not the best fit here and we can make it right at the first time.

backend/src/main/java/com/google/rolecall/ApplicationLoader.java
zephyr-l: This is already in a dedicated method. Please ignore 1.

backend/src/main/java/com/google/rolecall/services/PerformanceServices.java
zephyr-l: I do not think this is feature asked by AA. I am fine with the update but maybe we can also let them know as a new feature in the system.

backend/src/main/java/com/google/rolecall/ApplicationLoader.java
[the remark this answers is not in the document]
yhedholm: Done.

backend/src/main/java/com/google/rolecall/ApplicationLoader.java
yhedholm: Done.

backend/src/main/java/com/google/rolecall/services/PerformanceServices.java
yhedholm: It is actually a bug fix, fixing the bug I introduced when I overwrote a field that was needed for the casting logic. My first "fix" gave us ordered positions, but broke the casts. This fix makes the cast logic work. I don't think Alvin Ailey is aware of the bug, (it was only in the system for a week and it didn't show easily) so there isn't much need to talk about the change.

backend/src/main/java/com/google/rolecall/ApplicationLoader.java
zephyr-l: Thanks.

backend/src/main/java/com/google/rolecall/ApplicationLoader.java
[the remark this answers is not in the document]
zephyr-l: Thanks.

zephyr-l (approved):

PR #8, opened 2020-09-25
Frontend changes for revelation support

yhedholm:
Most of the changes to support 'Revelations' were made in the prior PR. To go 'all the way' included some fixes on the backend as well as the frontend. The largest changes on the frontend happened to the performance section. Those changes will come in a following PR.

The largest number of lines in this PR comes from 'cast-editor-v2.component.html where I 'narrowed' all lines while I made changes.

I feel most of the changes are straight forward, except possibly the ones in the first file: user_api.service.ts. We agreed that 'notifications' was less descriptive than 'canReceiveNotifications' so we changed this in the front end code. However, I didn't change it on the backend (as it would force a database wipe). As a result the backend and frontend didn't match and the data wasn't transferred, so I changed the backend interacting field back to notifications.

### DIFF
--- a/frontend/rolecall/src/app/cast/cast-drag-and-drop.component.ts
+++ b/frontend/rolecall/src/app/cast/cast-drag-and-drop.component.ts
@@ -190,7 +190,8 @@ export class CastDragAndDrop implements OnInit {
     this.castPositions = [];
     if (!this.castAPI.hasCast(this.selectedCastUUID)) {
       this.castSelected = false;
-      this.logging.logError("Couldn't find cast: " + this.selectedCastUUID);
+      // The below error means little. After any save the system loses tracf the uuid of the saved cast
+      // this.logging.logError("Couldn't find cast: " + this.selectedCastUUID);
       return;
     }
     this.cast = this.castAPI.castFromUUID(this.selectedCastUUID);
@@ -223,7 +224,6 @@ export class CastDragAndDrop implements OnInit {
       let maxDancerIx = 0;
       for (let groupIx = 0; groupIx < filledPos.groups.length; groupIx++) {
         const group = filledPos.groups[groupIx];
-        const group_index = group.group_index;
         for (let memberIx = 0; memberIx < group.members.length; memberIx++) {
           const member = group.members[memberIx];
           if (member.position_number >= maxDancerIx) {
@@ -235,7 +235,7 @@ export class CastDragAndDrop implements OnInit {
             }
           }
           const dancer = this.userAPI.users.get(member.uuid);
-          uiPos.castRows[member.position_number].subCastDancers[group_index] = {
+          uiPos.castRows[member.position_number].subCastDancers[group.group_index] = {
             uuid: dancer.uuid,
             firstName: dancer.first_name,
             lastName: dancer.last_name,
--- a/frontend/rolecall/src/app/cast/cast-drag-and-drop.component.ts
+++ b/frontend/rolecall/src/app/cast/cast-drag-and-drop.component.ts
@@ -190,8 +190,7 @@ export class CastDragAndDrop implements OnInit {
     this.castPositions = [];
     if (!this.castAPI.hasCast(this.selectedCastUUID)) {
       this.castSelected = false;
-      // The below error means little. After any save the system loses tracf the uuid of the saved cast
-      // this.logging.logError("Couldn't find cast: " + this.selectedCastUUID);
+      // TODO: After any save the system loses track the uuid of the saved cast. It would be nice to fix [YHE].
       return;
     }
     this.cast = this.castAPI.castFromUUID(this.selectedCastUUID);
--- a/frontend/rolecall/src/app/cast/cast-editor-v2.component.html
+++ b/frontend/rolecall/src/app/cast/cast-editor-v2.component.html
@@ -16,11 +16,11 @@
           </mat-icon>
           <div ngbDropdownMenu
               class="piece-options"
-              aria-labelledby="Piece Dropdown">
+              aria-label="Piece Dropdown">
             <button *ngFor="let piece of popupPieceList"
                 ngbDropdownItem
                [class.std-italic-font]="piece.hasSibling"
-               (click)="selectPiece(piece.pieceIx)">
+               (click)="selectPiece(piece.pieceIndex)">
               {{piece.name}}
             </button>
           </div>
--- a/frontend/rolecall/src/app/cast/cast-editor-v2.component.html
+++ b/frontend/rolecall/src/app/cast/cast-editor-v2.component.html
@@ -8,7 +8,7 @@
         <div ngbDropdownToggle
             class="app-card piece-select">
           <h1 class="left-card-title"
-              [ngStyle]="{'font-style': selectedPiece && selectedPiece.siblingId > 0 ? 'italic' : 'normal' }">
+              [class.std-italic-font]="selectedPiece && selectedPiece.siblingId > 0">
             {{selectedPiece ? selectedPiece.name : " "}}
           </h1>
           <mat-icon class="dropdownArrow">
@@ -17,10 +17,10 @@
           <div ngbDropdownMenu
               class="piece-options"
               aria-labelledby="Piece Dropdown">
-            <button *ngFor="let piece of popupList"
+            <button *ngFor="let piece of popupPieceList"
                 ngbDropdownItem
-                (click)="onSelectPiece(piece.pieceIx)"
-                [ngStyle]="{'font-style': piece.hasSibling ? 'italic' : 'normal' }">
+               [class.std-italic-font]="piece.hasSibling"
+               (click)="selectPiece(piece.pieceIx)">
               {{piece.name}}
             </button>
           </div>
@@ -29,10 +29,10 @@
       <div class="cast-select-container">
         <div *ngFor="let cast of filteredCasts"
             class="app-card cast-select"
-            [ngClass]="cast.uuid == castDragAndDrop.selectedCastUUID ? 'selected-cast' : ''"
+            [class.selected-cast]="cast.uuid == castDragAndDrop.selectedCastUUID"
             (click)="setCurrentCast(cast)">
           <h2 class="cast-name"
-              [ngStyle]="{'font-style': selectedPiece && selectedPiece.siblingId > 0 ? 'italic' : 'normal' }">
+              [class.std-italic-font]="selectedPiece && selectedPiece.siblingId > 0">
             {{ cast.name }}
           </h2>
         </div>
--- a/frontend/rolecall/src/app/cast/cast-editor-v2.component.html
+++ b/frontend/rolecall/src/app/cast/cast-editor-v2.component.html
@@ -1,34 +1,48 @@
 <app-loading-spinner *ngIf="!dataLoaded"></app-loading-spinner>
-<div [ngStyle]="dataLoaded ? {} : { 'display' : 'none' }" class="root-container">
+<div class="root-container"
+    [ngStyle]="dataLoaded ? {} : { 'display' : 'none' }">
   <div class="cast-selector-container">
     <div class="app-card card-margins left-card">
-      <div ngbDropdown #pieceDropdown="ngbDropdown">
-        <div class="app-card piece-select" ngbDropdownToggle>
-          <h1 class="left-card-title">
+      <div ngbDropdown
+          #pieceDropdown="ngbDropdown">
+        <div ngbDropdownToggle
+            class="app-card piece-select">
+          <h1 class="left-card-title"
+              [ngStyle]="{'font-style': selectedPiece && selectedPiece.siblingId > 0 ? 'italic' : 'normal' }">
             {{selectedPiece ? selectedPiece.name : " "}}
           </h1>
           <mat-icon class="dropdownArrow">
             arrow_drop_down
           </mat-icon>
-          <div ngbDropdownMenu aria-labelledby="Piece Dropdown" class="piece-options">
-            <button ngbDropdownItem *ngFor="let piece of allPieces" (click)="onSelectPiece(piece)">
+          <div ngbDropdownMenu
+              class="piece-options"
+              aria-labelledby="Piece Dropdown">
+            <button *ngFor="let piece of popupList"
+                ngbDropdownItem
+                (click)="onSelectPiece(piece.pieceIx)"
+                [ngStyle]="{'font-style': piece.hasSibling ? 'italic' : 'normal' }">
               {{piece.name}}
             </button>
           </div>
         </div>
       </div>
       <div class="cast-select-container">
-        <div *ngFor="let cast of filteredCasts" class="app-card cast-select"
-          [ngClass]="cast.uuid == castDragAndDrop.selectedCastUUID ? 'selected-cast' : ''"
-          (click)="setCurrentCast(cast)">
-          <h2 class="cast-name">
+        <div *ngFor="let cast of filteredCasts"
+            class="app-card cast-select"
+            [ngClass]="cast.uuid == castDragAndDrop.selectedCastUUID ? 'selected-cast' : ''"
+            (click)="setCurrentCast(cast)">
+          <h2 class="cast-name"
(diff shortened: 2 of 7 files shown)


frontend/rolecall/src/app/cast/cast-drag-and-drop.component.ts
```diff
@@ -190,7 +190,8 @@ export class CastDragAndDrop implements OnInit {
     this.castPositions = [];
     if (!this.castAPI.hasCast(this.selectedCastUUID)) {
       this.castSelected = false;
-      this.logging.logError("Couldn't find cast: " + this.selectedCastUUID);
+      // The below error means little. After any save the system loses tracf the uuid of the saved cast
```
mikzat, line 193: Ack, can remove the commented out code and corresponding comment.

frontend/rolecall/src/app/cast/cast-editor-v2.component.html
```diff
@@ -1,34 +1,48 @@
...
     <div class="app-card card-margins left-card">
-      <div ngbDropdown #pieceDropdown="ngbDropdown">
-        <div class="app-card piece-select" ngbDropdownToggle>
-          <h1 class="left-card-title">
+      <div ngbDropdown
+          #pieceDropdown="ngbDropdown">
+        <div ngbDropdownToggle
+            class="app-card piece-select">
+          <h1 class="left-card-title"
+              [ngStyle]="{'font-style': selectedPiece && selectedPiece.siblingId > 0 ? 'italic' : 'normal' }">
```
mikzat, line 11: This could be a little more readable as a dynamic class instead of `ngStyle`, something like:
`[class.italic-font]="selectedPiece && selectedPiece.siblingId > 0"`

frontend/rolecall/src/app/cast/cast-editor-v2.component.html
```diff
@@ -1,34 +1,48 @@
...
             {{selectedPiece ? selectedPiece.name : " "}}
           </h1>
           <mat-icon class="dropdownArrow">
             arrow_drop_down
           </mat-icon>
-          <div ngbDropdownMenu aria-labelledby="Piece Dropdown" class="piece-options">
-            <button ngbDropdownItem *ngFor="let piece of allPieces" (click)="onSelectPiece(piece)">
+          <div ngbDropdownMenu
+              class="piece-options"
+              aria-labelledby="Piece Dropdown">
```
mikzat, line 19: The value of an `aria-labelledby` must be an element ID. I think in this case the intention was for this to be an `aria-label`.

frontend/rolecall/src/app/cast/cast-editor-v2.component.html
```diff
@@ -1,34 +1,48 @@
...
+            <button *ngFor="let piece of popupList"
+                ngbDropdownItem
+                (click)="onSelectPiece(piece.pieceIx)"
```
mikzat, line 22: Can we rename `pieceIx` to `pieceIndex`, shortened words are not great for readability since not everyone has the same background of "norms"

frontend/rolecall/src/app/cast/cast-editor-v2.component.html
```diff
@@ -1,34 +1,48 @@
...
+                [ngStyle]="{'font-style': piece.hasSibling ? 'italic' : 'normal' }">
```
mikzat, line 23: Same here, if we had a `italic-font` class it can be used here dynamically as well
`[class.italic-font]="piece.hasSibling"`

frontend/rolecall/src/app/cast/cast-editor-v2.component.html
```diff
@@ -1,34 +1,48 @@
...
         </div>
       </div>
       <div class="cast-select-container">
-        <div *ngFor="let cast of filteredCasts" class="app-card cast-select"
-          [ngClass]="cast.uuid == castDragAndDrop.selectedCastUUID ? 'selected-cast' : ''"
-          (click)="setCurrentCast(cast)">
-          <h2 class="cast-name">
+        <div *ngFor="let cast of filteredCasts"
+            class="app-card cast-select"
+            [ngClass]="cast.uuid == castDragAndDrop.selectedCastUUID ? 'selected-cast' : ''"
```
mikzat, line 32: can be simplified
`[class.selected-cast]="cast.uuid === castDragAndDrop.selectedCastUUID"`

frontend/rolecall/src/app/cast/cast-editor-v2.component.ts
```diff
@@ -7,6 +7,13 @@ import { Cast, CastApi } from '../api/cast_api.service';
 import { Piece, PieceApi } from '../api/piece_api.service';
 import { CastDragAndDrop } from './cast-drag-and-drop.component';
 
+type PopupData = {
+  name: string;
+  sortString: string;
+  hasSibling: boolean;
+  pieceIx: number;
```
mikzat, line 14: pieceIndex

frontend/rolecall/src/app/cast/cast-editor-v2.component.ts
```diff
@@ -20,6 +27,7 @@ export class CastEditorV2 implements OnInit {
   allCasts: Cast[] = [];
   filteredCasts: Cast[] = [];
   allPieces: Piece[] = [];
+  popupList: PopupData[] = [];
```
mikzat, line 30: Let's call it something that describes what it represents rather than generic "popupData". How about:
`pieceMenuItems: PieceMenuItem[] = [];`

frontend/rolecall/src/app/cast/cast-editor-v2.component.ts
```diff
@@ -94,10 +102,26 @@ export class CastEditorV2 implements OnInit {
     }
   }
 
+  buildPopupList() {
+    this.popupList = this.allPieces.map((piece, pieceIx) => {
+      const existingCasts = this.allCasts.filter((val) => val.segment == piece.uuid);
```
mikzat, line 107: Rename `val` to `cast`

frontend/rolecall/src/app/cast/cast-editor-v2.component.ts
```diff
@@ -94,10 +102,26 @@ export class CastEditorV2 implements OnInit {
...
+      const prefix = existingCasts.length === 0 ? " " : "z";
```
mikzat, line 108: What does the "z" mean? Just for sorting? Feels like it would be better to not take a shortcut by reusing default alphabetical sort, and just add the sort logic inside the sort function below.

frontend/rolecall/src/app/cast/cast-editor-v2.component.ts
```diff
@@ -94,10 +102,26 @@ export class CastEditorV2 implements OnInit {
...
+      const name = existingCasts.length === 0 ? "*" + piece.name : piece.name;
```
mikzat, line 109: What does the "*" mean? If it's going to be displayed in the menu as part of the actual name then it's an accessibility issue (though not sure how much that's important to us right now)

frontend/rolecall/src/app/cast/cast-editor-v2.component.ts
```diff
@@ -94,10 +102,26 @@ export class CastEditorV2 implements OnInit {
     }
   }
 
+  buildPopupList() {
```
mikzat, line 105: make it private

frontend/rolecall/src/app/cast/cast-editor-v2.component.ts
```diff
@@ -55,8 +63,8 @@ export class CastEditorV2 implements OnInit {
     }
   }
 
-  onSelectPiece(piece: Piece) {
-    this.setPiece(piece);
+  onSelectPiece(pieceIx: number) {
```
mikzat, line 66: Would be nice to just call this `selectPiece`, doesn't need the "on" prefix

mikzat (changes requested):
Looks good overall, mostly minor naming things.

frontend/rolecall/src/app/cast/cast-editor-v2.component.ts
yhedholm: I use Ix for Index all over the place. Does it break a google rule?

mikzat: I noticed :) 

Here's the snippet from our internal requirements for var naming in JavaScript/TypeScript:
image

frontend/rolecall/src/app/cast/cast-editor-v2.component.ts
yhedholm: See the comment above. The problem was that the customer wanted an indication of which ballets had not been casted (or finished). Sorting these at the to and preceding the ballet name with a '*' seemed like a great improvement over what was before, and probably one that doesn't need to be changed going forward. (If we get accessibility issues, perhaps we can test for the leading *.)

frontend/rolecall/src/app/cast/cast-drag-and-drop.component.ts
yhedholm: I want to leave this comment in. I spent several hours tracking down this error and figuring out why it happened (and also realizing that it was very difficult to fix without some serious logic rewrite). The comment is marking where future work should happen.

mikzat: That makes sense, in that case let's remove the dead code line and just leave a comment in this format:
```
// TODO: After any save the system loses track the uuid of the saved cast, so we need to fix [XYZ].
```

frontend/rolecall/src/app/cast/cast-editor-v2.component.html
yhedholm: Yes. Sorry. Fixed.

frontend/rolecall/src/app/cast/cast-editor-v2.component.html
yhedholm: I already commented on this below. This would be a biggie. I don't have many conventions, but one is to use Ix for Index everywhere. I doubt anyone can read code of mine without rapidly understanding that Ix means Index. Am I breaking a hard google rule or is this more of a soft guideline?

mikzat: Actually the first time I read your PR with "Ix" my first thought was "intersection", not index, until I got quite a lot deeper into the code and started noticing more patterns which "clicked" for me.

Don't worry about refactoring the current usage if the impact is too large so it doesn't grow the current PR scope, just avoid using it in new places and I can help adjust the existing places separately.

frontend/rolecall/src/app/cast/cast-editor-v2.component.html
yhedholm: An interesting point. Personally, I prefer using [ngStyle] to [ngClass] or [clsss.xxx] because I can see right there what is happening without having to go and look it up in another file. If we can create some overarching structure to our scss files, I could be persuaded to switch to [class.xxx]. (I would like to have one central file with all 'standard style classes. These classes should have a standard prefix so they are easily identifiable, etc. If you agree that we should do that across the board, I can change this right now.)

mikzat: ngStyle can be useful in complex cases, but the motivation here is to reduce the amount of logic in templates especially for simple cases with a clear alternative.

Looks like there's currently already a global scss file called `styles.scss`, maybe this global class `italic-font` can go in there for now and maybe it'll evolve to a slightly different structure in the future if needed? Though let's not make any broader changes as part of the same PR so as not to increase its scope.

frontend/rolecall/src/app/cast/cast-editor-v2.component.ts
mikzat: Ack, I guess let's leave it as is for now and take another look later.

frontend/rolecall/src/app/cast/cast-editor-v2.component.html
yhedholm: After reading the similar comment below, I decided to start building a standard library of style classes.

Suggestion implemented.

frontend/rolecall/src/app/cast/cast-editor-v2.component.html
yhedholm: Done.

frontend/rolecall/src/app/cast/cast-editor-v2.component.html
yhedholm: Done.

frontend/rolecall/src/app/cast/cast-editor-v2.component.ts
yhedholm: Done.

frontend/rolecall/src/app/cast/cast-editor-v2.component.ts
yhedholm: Done

frontend/rolecall/src/app/cast/cast-editor-v2.component.ts
yhedholm: Yes, the "z" is just for sorting. I disagree that it is a shortcut. This list has alway been sorted alphabetically. The idea here is that every piece that is lacking a cast should be at the top of the list (and the name should preceded with a '*' to show that it hadn't been casted yet. Prepending " " and "z" to the old sort name is very easy to understand and requires a minimal amount of extra code, two goals of good code in my opinion.

frontend/rolecall/src/app/cast/cast-editor-v2.component.ts
yhedholm: Done.

frontend/rolecall/src/app/cast/cast-editor-v2.component.ts
yhedholm: OK.

frontend/rolecall/src/app/cast/cast-editor-v2.component.html
mikzat: Still looks the same `aria-labelledby`, maybe it got lost in the last commit

mikzat (changes requested):

frontend/rolecall/src/app/cast/cast-drag-and-drop.component.ts
yhedholm: Done.

frontend/rolecall/src/app/cast/cast-editor-v2.component.html
yhedholm: There were two of these. I fixed the other one earlier.

frontend/rolecall/src/app/cast/cast-editor-v2.component.html
yhedholm: OK.

frontend/rolecall/src/app/cast/cast-editor-v2.component.ts
yhedholm: I was hoping that Ix would considered as as valid as Id. Is the difference that Index is shorter than Identifier?

mikzat: I think difference is not about the character length, but more about what's considered "common knowledge" in computer science. "id" is globally known as a shorthand for "identifier", while "ix" is a personal convention not globally considered a standard.

(anecdotally, I've never seen ix used in my ~15yr career, but that's not really a deciding factor, just a data point)

frontend/rolecall/src/app/cast/cast-drag-and-drop.component.ts
mikzat: Perfect.

frontend/rolecall/src/app/cast/cast-editor-v2.component.html
mikzat: Ah I see, I missed the other one.

mikzat (approved):
Looks good!

PR #10, opened 2020-09-26
Cleaning up "user" frontend path.

mikzat:
There should be no functional changes from any of this. 

See description of each individual commit for details.

### DIFF
--- a/frontend/rolecall/src/app/user/user-editor.component.spec.ts
+++ b/frontend/rolecall/src/app/user/user-editor.component.spec.ts
@@ -1,40 +1,30 @@
-import {async, ComponentFixture, TestBed} from '@angular/core/testing';
+import {async, TestBed} from '@angular/core/testing';
 import {ActivatedRoute} from '@angular/router';
 import {EmptyStringIfUndefinedPipe} from '../common_components/empty_string_if_undefined.pipe';
 import {UserEditor} from './user-editor.component';
 
-let activatedRouteStub = {
-  snapshot: {
-    params: {
-      uuid: "testUUID"
-    }
-  }
-}
+const activatedRouteStub = {
+  snapshot: {params: {uuid: 'testUUID'}}
+};
 
 describe('UserEditorComponent', () => {
-  let component: UserEditor;
-  let fixture: ComponentFixture<UserEditor>;
+  let userEditor: UserEditor;
 
   beforeEach(async(() => {
     TestBed.configureTestingModule({
           declarations: [UserEditor, EmptyStringIfUndefinedPipe],
-          providers: [
-            {
-              provide: ActivatedRoute,
-              useValue: activatedRouteStub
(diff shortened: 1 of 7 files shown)


frontend/rolecall/src/app/user/user-editor.component.spec.ts
```diff
@@ -1,40 +1,30 @@
...
-      declarations: [UserEditor, EmptyStringIfUndefinedPipe],
-      providers: [
-        {
-          provide: ActivatedRoute,
-          useValue: activatedRouteStub
-        }
-      ]
-    })
-      .compileComponents();
+          declarations: [UserEditor, EmptyStringIfUndefinedPipe],
```
yhedholm, line 15: You did more than pretty this one up.

frontend/rolecall/src/app/user/user-editor.component.ts
```diff
@@ -1,121 +1,197 @@
...
       }
     }
   }
 
-  setCurrentUser(user: User, fromInputChange?: boolean, shouldSetLastUser?: boolean) {
+  setCurrentUser({user, fromInputChange, shouldSetLastUser}: {
+    user: User | undefined,
+    fromInputChange?: boolean,
+    shouldSetLastUser?: boolean,
+  }) {
```
yhedholm, line 157: Is this to force the caller to specify parameters?

frontend/rolecall/src/app/user/user-editor.component.ts
```diff
@@ -1,121 +1,197 @@
...
+      private readonly route: ActivatedRoute,
+      private readonly userAPI: UserApi,
+      private readonly location: Location) {
+  }
 
   ngOnInit(): void {
-    let uuid = this.route.snapshot.params.uuid;
-    if (!isNullOrUndefined(uuid)) {
+    const uuid: string = this.route.snapshot.params.uuid;
+    if (uuid) {
```
yhedholm, line 90: You don't always specify a type when assigning a const. Why here?

yhedholm (approved):
Thank you! I look forward to working with this new and improved file!

frontend/rolecall/src/app/user/user-editor.component.ts
mikzat: Good eye. The reason for changing it here is that `params` has the type `"[key: string]: any"` so the value of `uuid` evaluates to `"any"`. Narrowing the type as early as possible allows the rest of the code to know that it's actually a string for proper type checking.

frontend/rolecall/src/app/user/user-editor.component.ts
mikzat: Correct, this allows callers to be more readable as it's clear what parameters are being set. It also makes the arguments position-agnostic even though they are optional.

PR #13, opened 2020-09-29
Backend Changes for capturing middle name and suffix

yhedholm:
Alvin Ailey has several dancers who use their middle names actively (they are listed in programs etc.) and one dancer who has a suffix. In the past we could support this by adding extra names in the first name field and adding the suffix to the last name field.

With our new picture lookup logic, we can't edit either the first or last name fields any longer. Given that changing database structure creates access permission issues unless we wipe the database (which will be more difficult when we have active users with data), it makes sense to add the middle name and suffix fields now.

This code is very straight forward.

### DIFF
--- a/backend/src/main/java/com/google/rolecall/ApplicationLoader.java
+++ b/backend/src/main/java/com/google/rolecall/ApplicationLoader.java
@@ -54,7 +54,9 @@ public class ApplicationLoader implements ApplicationRunner {
     try {
       admin = User.newBuilder()
           .setFirstName(adminFirstName)
+          .setMiddleName("")
           .setLastName(adminLastName)
+          .setSuffix("")
           .setEmail(adminEmail)
           .setIsActive(true)
           .setCanLogin(true)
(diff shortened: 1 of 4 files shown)


backend/src/main/java/com/google/rolecall/ApplicationLoader.java
```diff
@@ -104,40 +108,40 @@ private void createOneUser(String fName, String lName, String email, String date
...
+      createOneUser("Ghrai", "", "Devore", "", "gde@gmail.com","1/1/2020");
+      createOneUser("Solomon", "", "Dumas", "", "sdu@gmail.com","1/1/2020");
+      createOneUser("Ronni", "", "Favors", "", "rf@gmail.com","1/1/2020");
+      createOneUser("Samantha", "", "Figgins", "", "sf@gmail.com","1/1/2020");
+
+      createOneUser("James", "", "Gilmer", "", "jg@gmail.com","1/1/2020");
+      createOneUser("Vernard", "", "Gilmore", "", "vg@gmail.com","1/1/2020");
+      createOneUser("Jacqueline", "", "Green", "", "jgr@gmail.com","1/1/2020");
+      createOneUser("Jacquelin", "", "Harris", "", "jh@gmail.com","1/1/2020");
+      createOneUser("Michael", "", "Jackson", " jr.", "mj@gmail.com","1/1/2020");
```
zephyr-l, line 126:
```suggestion
      createOneUser("Michael", "", "Jackson", " Jr.", "mj@gmail.com","1/1/2020");
```

zephyr-l (approved):

PR #15, opened 2020-09-29
Removed the old unused cast editor files

yhedholm:
I don't know why they left the old unused files in the project.


mikzat (approved):
Nice and quick :)

PR #17, opened 2020-09-29
style changes

yhedholm:
Style changes for the incremental cast count pr.


frontend/rolecall/src/app/cast/cast-drag-and-drop.component.scss
```diff
@@ -232,3 +235,114 @@ $cellHeight: 4rem;
   justify-content: space-between;
   width: 100%;
 }
+
+.button-h-wrapper {
```
mikzat, line 239: These classes are not used yet, let's leave them out and put them in later in the same PR where they are used, so that we don't have any unused code committed to 'master' branch at any point in time.

mikzat (changes requested):
Please resolve the comment before this one is ready to merge in

frontend/rolecall/src/app/cast/cast-drag-and-drop.component.scss
yhedholm: I get it now. You meant to remove the below lines. I thought you were talking about the whole file.

mikzat (approved):

PR #19, opened 2020-09-30
front end changes for middle name and suffix capture

yhedholm:
Most of the changed lines come from the changes in the mock data file.

There are two trivial changes on the backend (fixing a spelling error and creating a parallel database for easy testing).

Apart from the mock data changes, I am just adding two extra name field, wherever name fields are handled.

### DIFF
--- a/frontend/rolecall/angular.json
+++ b/frontend/rolecall/angular.json
@@ -29,8 +29,7 @@
             ],
             "styles": [
               "./node_modules/@angular/material/prebuilt-themes/deeppurple-amber.css",
-              "src/styles.scss",
-              "src/std.scss"
+              "src/styles.scss"
             ],
             "scripts": []
           },
@@ -95,8 +94,7 @@
             ],
             "styles": [
               "./node_modules/@angular/material/prebuilt-themes/deeppurple-amber.css",
-              "src/styles.scss",
-              "src/std.scss"
+              "src/styles.scss"
             ],
             "scripts": []
           }
--- a/frontend/rolecall/angular.json
+++ b/frontend/rolecall/angular.json
@@ -29,7 +29,8 @@
             ],
             "styles": [
               "./node_modules/@angular/material/prebuilt-themes/deeppurple-amber.css",
-              "src/styles.scss"
+              "src/styles.scss",
+              "src/std.scss"
             ],
             "scripts": []
           },
@@ -94,7 +95,8 @@
             ],
             "styles": [
               "./node_modules/@angular/material/prebuilt-themes/deeppurple-amber.css",
-              "src/styles.scss"
+              "src/styles.scss",
+              "src/std.scss"
             ],
             "scripts": []
           }
--- a/frontend/rolecall/src/app/performance/performance-editor.component.ts
+++ b/frontend/rolecall/src/app/performance/performance-editor.component.ts
@@ -691,7 +691,10 @@ export class PerformanceEditor implements OnInit, OnDestroy, AfterViewChecked {
               groups: val.groups.map(g => {
                 return {
                   ...g,
-                  memberNames: g.members.map(mem => this.userAPI.users.get(mem.uuid)).map(usr => usr.first_name + " " + usr.last_name)
+                  memberNames: g.members.map(mem => this.userAPI.users.get(mem.uuid)).map(
+                      usr => usr.first_name + " " +
+                      (usr.middle_name.length > 0 ? usr.middle_name + " " : "") +
+                      usr.last_name + usr.suffix),
                 }
               })
             }
--- a/frontend/rolecall/src/app/user/user-editor.component.html
+++ b/frontend/rolecall/src/app/user/user-editor.component.html
@@ -54,28 +54,28 @@
           <app-text-input
               valueName="First Name"
               (valueChange)="onInputChange($event)"
-              [initValue]="(currentSelectedUser | emptyStringIfUndefined)['first_name'] | emptyStringIfUndefined">
+              [initValue]="currentSelectedUser?.first_name || ''">
           </app-text-input>
         </div>
         <div class="name-input">
           <app-text-input
               valueName="Middle Name"
               (valueChange)="onInputChange($event)"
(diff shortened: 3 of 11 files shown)


frontend/rolecall/src/app/user/user-editor.component.html
```diff
@@ -12,8 +12,12 @@ <h1 class="left-card-title">
...
             (click)="setCurrentUser({user : user, fromInputChange : false, shouldSetLastUser : true})">
           <h2 class="user-name">
-            {{ (user.first_name ? user.first_name : " ") + " " + (user.last_name ? user.last_name
-              : " ") }}
+            {{
+              (user.first_name ? user.first_name + " " : "") + 
+              (user.middle_name ? user.middle_name + " " : "") +
+              (user.last_name ? user.last_name : "") + 
+              (user.suffix ? user.suffix : "") 
+            }}
```
mikzat, line 20: Missing space before suffix.

frontend/rolecall/src/app/user/user-editor.component.scss
```diff
@@ -1,5 +1,6 @@
 @import "src/colors";
 @import "src/constants";
+@import "src/std";
```
mikzat, line 3: Is this actually used anywhere?

frontend/rolecall/src/app/user/user-editor.component.html
```diff
@@ -12,8 +12,12 @@ <h1 class="left-card-title">
             [ngClass]="user.uuid == currentSelectedUser.uuid ? 'selected-user' : ''"
             (click)="setCurrentUser({user : user, fromInputChange : false, shouldSetLastUser : true})">
           <h2 class="user-name">
-            {{ (user.first_name ? user.first_name : " ") + " " + (user.last_name ? user.last_name
-              : " ") }}
+            {{
+              (user.first_name ? user.first_name + " " : "") +
```
mikzat, line 16: optional: would be nice to create a pure pipe that converts a user to a full name, then the same pipe can be used both in the template and in the class.

frontend/rolecall/src/app/performance/performance-editor.component.ts
```diff
@@ -691,7 +691,10 @@ export class PerformanceEditor implements OnInit, OnDestroy, AfterViewChecked {
               groups: val.groups.map(g => {
                 return {
                   ...g,
-                  memberNames: g.members.map(mem => this.userAPI.users.get(mem.uuid)).map(usr => usr.first_name + " " + usr.last_name)
+                  memberNames: g.members.map(mem => this.userAPI.users.get(mem.uuid)).map(
+                      usr => usr.first_name + " " +
+                      (usr.middle_name.length > 0 ? usr.middle_name + " " : "") +
+                      usr.last_name + usr.suffix),
```
mikzat, line 697: Missing space before suffix.

frontend/rolecall/src/app/user/user-editor.component.html
```diff
@@ -48,13 +57,27 @@ <h1 class="right-card-title">
               [initValue]="(currentSelectedUser | emptyStringIfUndefined)['first_name'] | emptyStringIfUndefined">
           </app-text-input>
         </div>
+        <div class="name-input">
+          <app-text-input
+              valueName="Middle Name"
+              (valueChange)="onInputChange($event)"
+              [initValue]="(currentSelectedUser | emptyStringIfUndefined)['middle_name'] | emptyStringIfUndefined">
```
mikzat, line 64: This will produce invalid code if `currentSelectedUser` is actually undefined, because it will try to take a substring from an empty string.

The proper thing to do would be to use null-safe property access:
```
[initValue]="currentSelectedUser?.middle_name || ''"
```

mikzat (changes requested):

frontend/rolecall/src/app/user/user-editor.component.html
yhedholm: Fixed.

frontend/rolecall/src/app/performance/performance-editor.component.ts
yhedholm: See comment above.

frontend/rolecall/src/app/user/user-editor.component.html
yhedholm: I like the idea! Let's do that as part of a later cleanup.

frontend/rolecall/src/app/user/user-editor.component.scss
yhedholm: Yes, this is where I put my standard style classes.

frontend/rolecall/src/app/user/user-editor.component.html
yhedholm: This is actually a 'feature'. The 'how to write a suffix' manuals are split on how to write John Smith junior. About half say it should be John Smith, Jr and the other half say it should be John Smith Jr. The easiest way to handle this is to give the user full control over the spacing. We discussed this with the Alving Ailey people and they liked this solution.

frontend/rolecall/src/app/user/user-editor.component.scss
mikzat: What I meant was I don't see anything from that file being used in `user-editor` styles, so it looks like it's imported unnecessarily?

yhedholm: I was going include stc/std in most scss files so that they are hooked up properly as we put more code into std.scss.

frontend/rolecall/angular.json
```diff
@@ -29,7 +29,8 @@
             ],
             "styles": [
               "./node_modules/@angular/material/prebuilt-themes/deeppurple-amber.css",
-              "src/styles.scss"
+              "src/styles.scss",
+              "src/std.scss"
```
mikzat, line 33: Revert the changes to this file and let's get https://github.com/googleinterns/role-call/pull/27 merged in.

mikzat (changes requested):

frontend/rolecall/angular.json
yhedholm: Done.

mikzat (approved):

PR #21, opened 2020-10-01
finish dynamic setting of #casts

yhedholm:
I hope this PR isn't too big. A lot of the diffs came from adjustments to the layout (making the cast table a little more compact)

The logic changes, approx. 20% of the code, are very straight forward. They basically replace a constant value CAST_COUNT with a variable castCount.


yhedholm:
Replaced by PR #25

PR #22, opened 2020-10-02
fixed casts loosing track of their uuid when saved

yhedholm:
When saved casts would lose track of their uuid, a lot of unnecessary and potentially dangerous code was executed to try to recover the uuid. Sometimes the user interface would be reset in the process.

At some point in the development process the patch save logic was changed so that during save, a cast would wipe its old database components before it would write new ones. Saving itself in a new position means a new database id which is the basis for the uuid.

I added a somewhat dirty way for the cast api to communicate the new id back to the savings routine.

In the process of debugging this, I found another problem: Every time a cast was saved it was followed by a read all cast call. This call executed a get all pieces call inside itself which would trigger a separate listener in the cast editor component. The only times pieces are needed in the cast editor is during cast creation and the casts always load their own pieces. Yet the cast editor, maybe for historical reasons, has its own onPievesLoad listener function.

The problem cased by this was that when the cast loader api got its pieces, the onPiecesLoader listener in the cast editor was triggered. Unfortunately, it was triggered too early so we didn't have the new cast collection yet. It caused a deadly embrace: when I updated the saved cast with the correct post-save uuid that worked with the new cast collection returned after the save, this new uuid clashed with a comparison with the old cast collection that was triggered by the get pieces listener that was triggered by the internal get all pieces call in the cast api.

Removing the onPiecesLoaded listener fixed that deadly embrace. Given that pieces are only used when new casts are created and the casts handle their own supply of pieces, removing the onPiecesLoaded listener should not be a problem. I found no problem during my own reasonably extensive testing.

### DIFF
--- a/frontend/rolecall/src/app/api/cast_api.service.ts
+++ b/frontend/rolecall/src/app/api/cast_api.service.ts
@@ -176,7 +176,7 @@ export class CastApi {
               let uniquePositionIDs = new Set<number>();
               rawCast.subCasts.forEach(val => uniquePositionIDs.add(val.positionId));
               return {
-                uuid: this.castUUIDFromRaw(rawCast.id),
+                uuid: CastApi.castUUIDFromRaw(rawCast.id),
                 name: rawCast.name,
                 segment: String(rawCast.sectionId),
                 castCount: highestCastNumber + 1,
@@ -417,7 +417,7 @@ async requestCastSet(cast: Cast): Promise<HttpResponse<any>> {
     return undefined;
   }
 
-  castUUIDFromRaw(id: number) {
+  static castUUIDFromRaw(id: number) {
     return String(id);
   }
 }
--- a/frontend/rolecall/src/app/api/cast_api.service.ts
+++ b/frontend/rolecall/src/app/api/cast_api.service.ts
@@ -176,7 +176,7 @@ export class CastApi {
               let uniquePositionIDs = new Set<number>();
               rawCast.subCasts.forEach(val => uniquePositionIDs.add(val.positionId));
               return {
-                uuid: CastApi.castUUIDFromRaw(rawCast.id),
+                uuid: CastApi.uuidFromRaw(rawCast.id),
                 name: rawCast.name,
                 segment: String(rawCast.sectionId),
                 castCount: highestCastNumber + 1,
@@ -417,7 +417,7 @@ async requestCastSet(cast: Cast): Promise<HttpResponse<any>> {
     return undefined;
   }
 
-  static castUUIDFromRaw(id: number) {
+  static uuidFromRaw(id: number) {
     return String(id);
   }
 }
--- a/frontend/rolecall/src/app/api/cast_api.service.ts
+++ b/frontend/rolecall/src/app/api/cast_api.service.ts
@@ -64,6 +64,7 @@ export type Cast = {
   filled_positions: CastPosition[];
 }
 
+// Get all casts
 export type AllCastsResponse = {
   data: {
     casts: Cast[]
@@ -71,6 +72,7 @@ export type AllCastsResponse = {
   warnings: string[]
 };
 
+// Get one specific cast
 export type OneCastResponse = {
   data: {
     cast: Cast
@@ -176,7 +178,7 @@ export class CastApi {
               let uniquePositionIDs = new Set<number>();
               rawCast.subCasts.forEach(val => uniquePositionIDs.add(val.positionId));
               return {
-                uuid: CastApi.uuidFromRaw(rawCast.id),
+                uuid: String(rawCast.id),
                 name: rawCast.name,
                 segment: String(rawCast.sectionId),
                 castCount: highestCastNumber + 1,
@@ -213,8 +215,8 @@ async requestCastSet(cast: Cast): Promise<HttpResponse<any>> {
       observe: "response",
       withCredentials: true
     }).toPromise().catch((errorResp) => errorResp).then(
-        resp => this.respHandler.checkResponse<any>(resp)).then(async na => {
-      const rawCast = this.PatchPostPrep(cast);
+        resp => this.respHandler.checkResponse<any>(resp)).then(async notUsed => {
+      const rawCast = this.patchPostPrep(cast);
       return this.http.post(environment.backendURL + "api/cast", rawCast, {
         headers: header,
         observe: "response",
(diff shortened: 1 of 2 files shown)


frontend/rolecall/src/app/api/cast_api.service.ts
```diff
@@ -417,7 +417,7 @@ async requestCastSet(cast: Cast): Promise<HttpResponse<any>> {
     return undefined;
   }
 
-  static castUUIDFromRaw(id: number) {
+  static uuidFromRaw(id: number) {
```
mikzat, line 420: This util doesn't really have much to do with UUID or the CastAPI class. If it's really needed, maybe it can be just a generic global util
```
export function numToString(num: number) {
  return String(num);
}
```

Though I wonder how useful that is, as it's simpler to just write 
`String(this.castAPI.lastSavedCastId)`
instead of 
`numToString(this.castAPI.lastSavedCastId`
`CastApi.uuidFromRaw(this.castAPI.lastSavedCastId)`

frontend/rolecall/src/app/api/cast_api.service.ts
```diff
@@ -35,6 +35,11 @@ type AllRawCastsResponse = {
   warnings: string[]
 }
 
+type OneRawCastsResponse = {
```
mikzat, line 38: Please a comment about what this means, or could we pick a better name? Not clear what "one" means here

frontend/rolecall/src/app/api/cast_api.service.ts
```diff
@@ -169,7 +176,7 @@ export class CastApi {
               let uniquePositionIDs = new Set<number>();
               rawCast.subCasts.forEach(val => uniquePositionIDs.add(val.positionId));
               return {
-                uuid: String(rawCast.id),
+                uuid: CastApi.uuidFromRaw(rawCast.id),
```
mikzat, line 179: I think the previous approach of `uuid: String(rawCast.id),` was better than adding this indirection since the conversion doesn't require any additional logic.

frontend/rolecall/src/app/api/cast_api.service.ts
```diff
@@ -192,95 +199,73 @@ export class CastApi {
...
-  async requestCastSet(cast: Cast): Promise<HttpResponse<any>> {
-    if (environment.mockBackend) {
-      return this.mockBackend.requestCastSet(cast);
-    }
-    // Check if we have record of the cast and patch if we do
-    if (this.hasCast(cast.uuid)) {
-      // Do patch
-      let header = await this.headerUtil.generateHeader();
-      return this.http.delete(environment.backendURL + 'api/cast?castid=' + cast.uuid, {
+/** Hits backend with create/edit cast POST request */
```
mikzat, line 202: Was the indentation amount changed here?

frontend/rolecall/src/app/api/cast_api.service.ts
```diff
@@ -192,95 +199,73 @@ export class CastApi {
...
+    return this.http.post(environment.backendURL + "api/cast", rawCast, {
+      headers: header,
+      observe: "response",
+      withCredentials: true
+    }).toPromise().catch((errorResp) => errorResp).then(
+        resp => this.respHandler.checkResponse<any>(resp));
+  }
+}
+
+  private PatchPostPrep(cast: Cast): RawCast {
```
mikzat, line 238: Always use camelCase for function names rather than UpperCammelCase

frontend/rolecall/src/app/api/cast_api.service.ts
```diff
@@ -192,95 +199,73 @@ export class CastApi {
...
+  // Check if we have record of the cast and patch if we do
+  if (this.hasCast(cast.uuid)) {
+    // Do patch
+    let header = await this.headerUtil.generateHeader();
+    return this.http.delete(environment.backendURL + 'api/cast?castid=' + cast.uuid, {
+      headers: header,
+      observe: "response",
+      withCredentials: true
+    }).toPromise().catch((errorResp) => errorResp).then(
+        resp => this.respHandler.checkResponse<any>(resp)).then(async na => {
```
mikzat, line 216: Do you know why the response variable is called `na`? It looks like it's not used.

mikzat (changes requested):
I know this is existing code, but it's pretty hard to follow the architecture here. Since I don't have a full picture yet, it's hard to comment on the solution, so just trying to point out the few things that I do understand :)

frontend/rolecall/src/app/api/cast_api.service.ts
yhedholm: I assume it means not used / not applicable. I changed it to notUsed.

frontend/rolecall/src/app/api/cast_api.service.ts
yhedholm: Fixed. Sorry.

frontend/rolecall/src/app/api/cast_api.service.ts
yhedholm: I broke out completely duplicated code it the two branches ('Do patch' and 'Do post'). The remaining code indentation level may have changed as a result.

frontend/rolecall/src/app/api/cast_api.service.ts
yhedholm: Done.

frontend/rolecall/src/app/api/cast_api.service.ts
yhedholm: Done.

frontend/rolecall/src/app/api/cast_api.service.ts
yhedholm: Removed.

frontend/rolecall/src/app/api/cast_api.service.ts
mikzat: This line is still missing a comment

frontend/rolecall/src/app/api/cast_api.service.ts
mikzat: That's ok, but could you please fix the indent on this `requestCastSet()` function? Currently diff is showing every line as changed when it actually was not.

image

mikzat (changes requested):

mikzat (approved):

PR #23, opened 2020-10-02
Mostly backend changes for new segment navigation

yhedholm:
The customer was confused by our current segment navigation. Basically, once we added support for revelations like super ballets, where there exists a hierarchy between super ballets and the ballets that belong to them, a linear interface is inadequate, in spite of the tweaks we added to identify the different types of pieces/segments.

The customer asked for a folder like structure where super ballets can be open, and expose their ballets, or closed, and hide them. This requires a database change, and that is why I am rushing this change.

The code is pretty straight forward: Add a isOpen Boolean field to the piece/section table.


yhedholm:
After analyzing the multiuser implication of the database storing the display state of a Super Ballet, I decided to remove the database changes. The remaining changes now have to do with changing the 'REVELATION' code to 'SUPER'.

frontend/rolecall/src/app/piece/piece_editor.component.ts
```diff
@@ -232,6 +233,7 @@ export class PieceEditor implements OnInit {
     this.prevWorkingState = undefined;
     const newPiece: WorkingPiece = {
       uuid: 'segment:' + Date.now(),
+      isOpen: type === 'REVELATION' ? true : false,
```
mikzat, line 236: Please add a comment for why this should be true when type is REVELATION

mikzat (changes requested):

frontend/rolecall/src/app/piece/piece_editor.component.ts
yhedholm: Done.

mikzat (approved):

zephyr-l (approved):

PR #24, opened 2020-10-02
name change only: REVELATION -> SUPER

yhedholm:
We thought REVELATION was a type, but it is only one of Mr Ailey's Super Ballets. The Alvin Ailey people prefer the type name 'Super Ballets'. We will use SUPER as the internal type name.


frontend/rolecall/src/app/performance/performance-editor.component.ts
```diff
@@ -350,7 +350,7 @@ export class PerformanceEditor implements OnInit, OnDestroy, AfterViewChecked {
 
   // Step 2 -------------------------------------------------------
 
-  hasRevelation: boolean;
+  hasSuper: boolean;
```
mikzat, line 353: I think it's fine for the type to be `SUPER`, but the var names and function names it could be more clear as `SuperBallet`, for example `hasSuperBallet`.

mikzat (approved):

PR #25, opened 2020-10-03
basted on the last master, a minimum diff PR for finishing dynamic setting of #casts.

yhedholm:
This is the smallest PR I could make to finish the dynamic setting of cast counts.

I have closed the other PRs relating to this.

### DIFF
--- a/frontend/rolecall/src/app/cast/cast-drag-and-drop.component.html
+++ b/frontend/rolecall/src/app/cast/cast-drag-and-drop.component.html
@@ -27,60 +27,106 @@
     <div
         *ngFor="let position of castPositions; let i1 = index"
         class="position-container">
-      <div *ngIf="i1 < castPositions.length">
         <div class="table-container">
           <div
               class="horizontal-list"
               id="col-headers">
             <div class="cell cellBothHeader">
-              {{ castPositions[i1].pos.name }}
-            </div>
-            <div class="dancer-count-buttons-container">
-              <button
-                  mat-icon-button
-                  aria-label="Decrement Dancer Count Button"
-                  class="dancer-count-button"
-                  [disabled]="position.castSize < 2"
-                  (click)="decrementDancerCount(i1)">
-                <mat-icon [inline]="true"
-                    class="dancer-count-icon">
-                  remove_circle_outline
-                </mat-icon>
-              </button>
-              <button
-                  mat-icon-button
-                  aria-label="Increment Dancer Count Button"
-                  class="dancer-count-button"
-                  (click)="incrementDancerCount(i1)">
-                <mat-icon [inline]="true"
-                    class="dancer-count-icon">
-                  add_circle_outline
-                </mat-icon>
-              </button>
+              <div class="double-header-cell">
+                <div class="double-header-text">
+                  {{ castPositions[i1].pos.name }}
+                </div>
+                <div class="double-header-buttons">
+                  <div class="button-v-wrapper">
+                    <div class="button-v-container">
+                      <button
+                          mat-icon-button
+                          aria-label="Decrement Dancer Count Button"
+                          class="count-button1"
+                          [disabled]="position.dancerCount < 2"
+                          (click)="decrementDancerCount(i1)">
+                        <mat-icon
+                            [inline]="true"
+                            class="count-icon1">
+                          remove_circle_outline
+                        </mat-icon>
+                      </button>
+                      <button
+                          mat-icon-button
+                          aria-label="Increment Dancer Count Button"
+                          class="count-button2"
+                          (click)="incrementDancerCount(i1)">
+                        <mat-icon
+                            [inline]="true"
+                            class="count-icon2">
+                          add_circle_outline
+                        </mat-icon>
+                      </button>
+                    </div>
+                  </div>  
+                </div>
+              </div>  
             </div>
             <div
-                *ngFor="let header of subCastHeaders"
-                class="cell cellColHeader">
+                *ngFor="let header of subCastHeaders; let i2 = index"
+                class="cell cellColHeader"
+                [class.adjust-last-cast-cell]="i2 === subCastHeaders.length - 1">
+              <ng-container
+                  *ngIf="i2 < subCastHeaders.length - 1; else lastCast">
                 {{ header }}
+              </ng-container>
+              <ng-template #lastCast>
+                <div class="last-cast-cell">
+                  <div class="last-cast-text">
+                    {{ header }}
+                  </div>
+                  <div class="button-h-wrapper">
+                    <div class="button-h-container">
+                      <button
+                          mat-icon-button
+                          aria-label="Decrement Dancer Count Button"
+                          class="count-button1"
+                          [disabled]="subCastHeaders.length < 2"
+                          (click)="decrementCastCount()">
+                        <mat-icon [inline]="true"
+                            class="count-icon1">
+                          remove_circle_outline
+                        </mat-icon>
+                      </button>
+                      <button
+                          mat-icon-button
+                          aria-label="Increment Dancer Count Button"
+                          class="count-button2"
+                          (click)="incrementCastCount()">
+                        <mat-icon
+                            [inline]="true"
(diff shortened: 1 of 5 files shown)


frontend/rolecall/src/app/cast/cast-drag-and-drop.component.html
```diff
@@ -27,60 +27,106 @@
...
+                *ngFor="let header of subCastHeaders; let i2 = index"
+                class="cell cellColHeader"
+                [class.adjust-last-cast-cell]="i2 === subCastHeaders.length - 1">
+              <ng-container
+                  *ngIf="i2 < subCastHeaders.length - 1; else lastCast">
                 {{ header }}
+              </ng-container>
+              <ng-template #lastCast>
+                <div class="last-cast-cell">
+                  <div class="last-cast-text">
```
mikzat, line 80: Please use `<p>` or other appropriate semantic elements for text, for better accessibility we should almost never use `<div>` or `<span>` purely for containing text (with some very specific exceptions).

frontend/rolecall/src/app/cast/cast-drag-and-drop.component.html
```diff
@@ -27,60 +27,106 @@
...
+                    {{ header }}
+                  </div>
+                  <div class="button-h-wrapper">
+                    <div class="button-h-container">
```
mikzat, line 84: Why do we need a div inside a div? `<div>`s that only contain another `<div>` should probably all be flattened to just one `<div>` layer with multiple classes if needed. Otherwise it makes the HTML more messy, forces browser to generate unnecessary elements, and also hurts accessibility.

frontend/rolecall/src/app/cast/cast-drag-and-drop.component.ts
```diff
@@ -21,7 +21,7 @@ type UICastRow = {
 
 type UICastPosition = {
   pos: Position;
-  castSize: number;
+  dancerCount: number;
```
mikzat, line 24: Is the intention to rename `UICastPosition` to `UIDancerPosition` and `castRows: UICastRow[]` to `dancerRows: UIDancerRow[]`?

Or is there an intentional difference here between a `cast` and a `dancer`?

frontend/rolecall/src/app/cast/cast-drag-and-drop.component.ts
```diff
@@ -103,6 +100,16 @@ export class CastDragAndDrop implements OnInit {
     this.castChangeEmitter.emit(this.cast);
   }
 
+  private buildSubCastHeader() {
```
mikzat, line 103: good extraction

frontend/rolecall/src/app/cast/cast-drag-and-drop.component.ts
```diff
@@ -343,13 +355,49 @@ export class CastDragAndDrop implements OnInit {
...
   incrementDancerCount(positionIndex: number) {
     this.castPositions[positionIndex].castRows.push({
-      subCastDancers: new Array(CAST_COUNT),
+      subCastDancers: new Array(this.castCount),
     });
-    this.castPositions[positionIndex].castSize += 1;
+    this.castPositions[positionIndex].dancerCount += 1;
+  }
+
+  private changeCastCount(change: number) {
```
mikzat, line 368: Currently it sounds like it will SET the count to the given value rather than increment/decrement, I recommend this for better readability:
```
/** Increments cast count by given amount, pass in negative value to decrement. */
private incrementCastCount(amount: number) {
```

frontend/rolecall/src/app/cast/cast-drag-and-drop.component.ts
```diff
@@ -343,13 +355,49 @@ export class CastDragAndDrop implements OnInit {
...
+      const newPosition: UICastPosition = {
+        pos: oldPosition.pos,
+        dancerCount: oldPosition.dancerCount,
+        castRows: [],
+      };
+      for(let dancerIndex = 0; dancerIndex < oldPosition.castRows.length; dancerIndex++) {
+        const oldCastRow = oldPosition.castRows[dancerIndex];
+        const transferCount = Math.min(this.castCount, oldCastCount)
+        const castRow: UICastRow = {
+          subCastDancers: new Array(this.castCount) as UICastDancer[],
```
mikzat, line 384: Let's try to avoid using `as` to force casting. It's dangerous since it overrides the type-check system, which means if the logic is refactored in the future and is wrong, it will not be caught by compiler and will result in runtime exception.

Try this:
```
subCastDancers: Array.from<UICastDancer>({length: this.castCount}),
```

mikzat (changes requested):

frontend/rolecall/src/app/cast/cast-drag-and-drop.component.html
yhedholm: Done.

frontend/rolecall/src/app/cast/cast-drag-and-drop.component.html
yhedholm: Done.

frontend/rolecall/src/app/cast/cast-drag-and-drop.component.ts
yhedholm: No, the purpose is to clarify an ambiguity. In the past, the number of casts was was a fluid number that was stored indirectly (dancers were stored in casts and the # of casts was deduced by in what casts the dancers had been stored.

When I changed the cast logic, I used castSize to specify how many dancers the cast had for a certain position. However, when the customer wanted to be able to change # of casts dynamically, we needed to keep track of #casts, so I added the field castCount. castCount and castSize was confusing (I confused them myself more than once) so I felt that changing castSize to dancerCount, which it really was, was easier to pair with castCount than the old castSize.

In the process of changing the cast logic, I changed how the cast is stored in memory. In the past there was a closer fit between how casts were stored in the database and how they were stored in memory than we have now. It is something that we may want to revisit when we clean up the code.

frontend/rolecall/src/app/cast/cast-drag-and-drop.component.ts
yhedholm: Done.

frontend/rolecall/src/app/cast/cast-drag-and-drop.component.ts
yhedholm: Done.

frontend/rolecall/src/app/cast/cast-drag-and-drop.component.html
mikzat: Still more work to do on this in the file, but don't want to hold up this PR.

frontend/rolecall/src/app/cast/cast-drag-and-drop.component.ts
mikzat: As a non-subject-matter-expert, to me "dancerCount" sounds exactly the same as "castSize" and "castCount", so it just looks like a typo that `UICastPosition` has `castRows` but `dancerCount`. 

But if you think it's more clear, we don't need to do anything else here right now.

frontend/rolecall/src/app/cast/cast-drag-and-drop.component.ts
```diff
@@ -365,9 +365,10 @@ export class CastDragAndDrop implements OnInit {
     this.castPositions[positionIndex].dancerCount += 1;
   }
 
-  private changeCastCount(change: number) {
+/** Increments cast count by given amount, pass in negative value to decrement. */
+private incrementCastCountUtil(amount: number) {
```
mikzat, line 369: Why the 'util" suffix? It seems redundant.

mikzat (approved):

frontend/rolecall/src/app/cast/cast-drag-and-drop.component.ts
yhedholm: There is already and incrementCastCount function in the system.

PR #26, opened 2020-10-03
the frontend changes for the new segment navigation

yhedholm:
The customer was confused by our current segment navigation. Basically, once we added support for revelations like super ballets, where there exists a hierarchy between super ballets and the ballets that belong to them, a linear interface is inadequate, in spite of the tweaks we added to identify the different types of pieces/segments.

The customer asked for a folder like structure where super ballets can be open, and expose their ballets, or closed, and hide them. This requires a database change, and that is why I am rushing this change.

### DIFF
--- a/frontend/rolecall/src/app/api/piece_api.service.ts
+++ b/frontend/rolecall/src/app/api/piece_api.service.ts
@@ -288,4 +288,8 @@ export class PieceApi {
       }
     });
   }
+
+  static uuidFromRaw(id: number) {
+    return String(id);
+  }
 }
--- a/frontend/rolecall/src/app/api/piece_api.service.ts
+++ b/frontend/rolecall/src/app/api/piece_api.service.ts
@@ -288,8 +288,4 @@ export class PieceApi {
       }
     });
   }
-
-  static uuidFromRaw(id: number) {
-    return String(id);
-  }
 }
--- a/frontend/rolecall/src/app/piece/piece_editor.component.ts
+++ b/frontend/rolecall/src/app/piece/piece_editor.component.ts
@@ -113,13 +113,13 @@ export class PieceEditor implements OnInit {
             this.buildRenderingItem(displayPiece, displayieceIndex));
   }
 
-  private buildRenderingItem(displayPiece: WorkingPiece, displayieceIndex: number) {
+  private buildRenderingItem(displayPiece: WorkingPiece, displayPieceIndex: number) {
     const hasNoChidren = displayPiece.type === 'SEGMENT'
         ? false : displayPiece.positions.length === 0;
     const name = hasNoChidren ? '*' + displayPiece.name : displayPiece.name;
     return {
       name,
-      pieceIndex: displayieceIndex,
+      pieceIndex: displayPieceIndex,
       siblingId: displayPiece.siblingId,
       type: displayPiece.type,
       isOpen: displayPiece.isOpen,
--- a/frontend/rolecall/src/app/piece/piece_editor.component.ts
+++ b/frontend/rolecall/src/app/piece/piece_editor.component.ts
@@ -26,14 +26,15 @@ type DraggablePosition = {
 type WorkingPiece = Piece & {
   addingPositions: DraggablePosition[];
   originalName: string;
+  isOpen;
 };
 
 type RenderingItem = {
   name: string;
-  sortString: string;
   pieceIndex: number;
   siblingId: number;
   type: PieceType;
+  isOpen: boolean;
   uuid: string
 };
 
@@ -48,6 +49,8 @@ export class PieceEditor implements OnInit {
 
   // All pieces or segments in the system ready to be edited.
   workingPieces: WorkingPiece[];
+  // Displayed items (some children of Super Ballets may be hidden)
+  displayedPieces:  WorkingPiece[];
   // List of segments visible on the left side of the page.
   renderingItems: RenderingItem[];
   urlPointingUUID: string;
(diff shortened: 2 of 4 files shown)


frontend/rolecall/src/app/api/piece_api.service.ts
```diff
@@ -288,4 +288,8 @@ export class PieceApi {
       }
     });
   }
+
+  static uuidFromRaw(id: number) {
```
mikzat, line 292: No need to use static since it's a basic util, just pull it out of the class and export it as a simple function. Will be easier to refactor later when we have a bunch of them and they need to extract to a separate file.

frontend/rolecall/src/app/piece/piece_editor.component.ts
```diff
@@ -26,14 +26,15 @@ type DraggablePosition = {
 type WorkingPiece = Piece & {
   addingPositions: DraggablePosition[];
   originalName: string;
+  isOpen;
```
mikzat, line 29: There's no type on `isOpen`, should there be?

frontend/rolecall/src/app/piece/piece_editor.component.ts
```diff
@@ -88,22 +91,37 @@ export class PieceEditor implements OnInit {
   }
 
   private buildRenderingList() {
-    this.renderingItems = this.workingPieces.map(
-        (workingPiece, workingPieceIndex) => {
-      const hasNoChidren = workingPiece.type === 'SEGMENT'
-          ? false : workingPiece.positions.length === 0;
-      const sortPrefix = hasNoChidren ? ' ' : 'z';
-      const name = hasNoChidren ? '*' + workingPiece.name : workingPiece.name;
+    // Remove Super Ballet children
```
mikzat, line 94: All the new logic that's building `displayedPieces` looks like it can be extracted to a separate function.

frontend/rolecall/src/app/piece/piece_editor.component.ts
```diff
@@ -88,22 +91,37 @@ export class PieceEditor implements OnInit {
...
+    this.displayedPieces = this.workingPieces.filter(piece => !piece.siblingId);
+    this.displayedPieces.sort((a, b) => a.name < b.name ? -1 : 1);
```
mikzat, line 96: Hmm, it's too bad that we have to do all this work manually to figure out what is visible and what isn't. 

Is there some existing UI component we can reuse managing layout and visibility state? I'm not familiar with the component library we are using here, but I know the Angular Material components have [expansion panels](https://material.angular.io/components/expansion/overview) and [trees](https://material.angular.io/components/tree/overview), those would be the first things I would look for if I was to implement similar functionality (if something similar is available).

frontend/rolecall/src/app/piece/piece_editor.component.ts
```diff
@@ -526,4 +545,13 @@ export class PieceEditor implements OnInit {
     }
     return code;
   }
+
+  toggleOpen(index: number) {
+    const superBallet = this.displayedPieces[index];
+    if (superBallet.type == 'SUPER') {
+      superBallet.isOpen = !superBallet.isOpen;
+      this.pieceAPI.setPiece(superBallet);
```
mikzat, line 553: Why does visibility state need to be managed on the API layer? It seems like it's a presentational concern that the API layer shouldn't know about. Also this is not something that should be sent to the backend as it will slow down UI interactions.

mikzat (changes requested):

frontend/rolecall/src/app/api/piece_api.service.ts
yhedholm: You are right that there is no need right now. However, the id -> uuid transformation could easily be much more involved. If that case it is also likely that different classes would have different logic for going from a database id to a frontend uuid. I though it was a good idea to make every class have a static function that you can call every time you need to generate a uuid from a database id.

yhedholm: Removed.

frontend/rolecall/src/app/piece/piece_editor.component.ts
yhedholm: Yes it is a boolean. Sorry!

frontend/rolecall/src/app/piece/piece_editor.component.ts
yhedholm: Done.

frontend/rolecall/src/app/piece/piece_editor.component.ts
yhedholm: The hiding and closing of the 'Super Ballet' children is the easy part. The tricky part is to show them in the right order etc. (For instance, the overall list is supported to be sorted in alphabetic order, while the children of the Super Ballet are supposed to be sorted in performance order.) Therefore I believe we would have to write most of this code again, even if we had an expansion panel. In our case, we are purposefully reusing the exact existing display object. The logic is not really to figure out how display and hide, the logic is figuring out which objects should be shown or hidden.

frontend/rolecall/src/app/piece/piece_editor.component.ts
yhedholm: This has already been removed from the sibling PR.

frontend/rolecall/src/app/piece/piece_editor.component.ts
mikzat: We can revisit this one in the future to see if it's worth looking into.

frontend/rolecall/src/app/piece/piece_editor.component.ts
```diff
@@ -101,27 +101,30 @@ export class PieceEditor implements OnInit {
...
-      return {
-        name,
-        pieceIndex: displayieceIndex,
-        siblingId: displayPiece.siblingId,
-        type: displayPiece.type,
-        isOpen: displayPiece.isOpen,
-        uuid: displayPiece.uuid,
-      };
-    });
+        (displayPiece, displayieceIndex) =>
```
mikzat, line 112: Spelling: looks like 4 usages of `displayieceIndex`  should be `displayPieceIndex`

mikzat (changes requested):

mikzat (approved):

PR #27, opened 2020-10-05
Fix global styles to be included from a single location.

mikzat:


yhedholm (approved):

yhedholm:
I am a little disappointed that we had to use !important.

PR #28, opened 2020-10-06
Backend cleanup

yhedholm:
1) Alvin Ailey asked us to add some preprogrammed users to the user list. They wanted them to have admin privileges so I had to add a parameter to the temporary createUser routine.

2) Some of the already approved 'REVELATION' to 'SUPER' changes got lost (perhaps because we had so many overlapping PRs). I put them back


yhedholm:
This PR was missing one merge and was replaced with PR 29

PR #29, opened 2020-10-06
backend cleanup

yhedholm:
Alvin Ailey asked us to add some preprogrammed users to the user list. They wanted them to have admin privileges so I had to add a parameter to the temporary createUser routine.

### DIFF
--- a/backend/src/main/java/com/google/rolecall/ApplicationLoader.java
+++ b/backend/src/main/java/com/google/rolecall/ApplicationLoader.java
@@ -136,7 +136,7 @@ public class ApplicationLoader implements ApplicationRunner {
       createOneUser("Renaldo", "", "Maurice", "", "rm@gmail.com","1/1/2020", false);
       createOneUser("Corrin", "Rachelle", "Mitchell", "", "crm@gmail.com","1/1/2020", false);
       createOneUser("Chalvar", "", "Monteiro", "", "cm@gmail.com","1/1/2020", false);
-      // BelÃ©n | Pereyra-Alem
+      // Belén | Pereyra-Alem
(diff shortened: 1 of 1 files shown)


backend/src/main/java/com/google/rolecall/ApplicationLoader.java
```diff
@@ -98,50 +98,54 @@ private void createOneUser(String fName, String mName, String lName, String suff
...
-      createOneUser("Miranda", "", "Quinn", "", "mq@gmail.com","1/1/2020");
-      createOneUser("Jamar", "", "Roberts", "", "jr@gmail.com","1/1/2020");
-      createOneUser("Matthew", "", "Rushing", "", "mr@gmail.com","1/1/2020");
-      createOneUser("Kanji", "", "Segawa", "", "ks@gmail.com","1/1/2020");
-      createOneUser("Courtney", "Celeste", "Spears", "", "ccs@gmail.com","1/1/2020");
-      createOneUser("Jermaine", "", "Terry", "", "jt@gmail.com","1/1/2020");
-      createOneUser("Christopher", "", "Wilson", "", "cw@gmail.com","1/1/2020");
-
-      createOneUser("Brandon", "", "Woolridge", "", "bw@gmail.com","1/1/2020");
+      createOneUser("Robert", "", "Battle", "", "rb@gmail.com","1/1/2020", true);
```
zephyr-l, line 112: I have concerns about submit these email to the public repo. Please reconsider.

mikzat: That's a great point, in fact these are already submitted. This might be a privacy violation by posting all users emails publicly.

PR #30, opened 2020-10-06
Fix color import

mikzat:
Moved styles to dedicated folder and adjusted global style providers and paths.

For some reason, cannot import "color" using absolute path, possibly a conflict with some other dependency? Either way, using relative path for colors works.


yhedholm (commented):
It would have been nice to get rid of the !important. Other than that, everything works fine.

yhedholm (approved):

PR #31, opened 2020-10-07
fix two bugs and one indentation error

yhedholm:
This PR fixes two bugs and one big indentation error (which is big enough to confuse the diff. No logic changed but the diff is not so sure).

Bug one was fixed in a prior PR but the final implementation was lost.

Bug two causes the system to misbehave with empty segment lists.

### DIFF
--- a/frontend/rolecall/src/app/piece/piece_editor.component.ts
+++ b/frontend/rolecall/src/app/piece/piece_editor.component.ts
@@ -68,7 +68,7 @@ export class PieceEditor implements OnInit {
 
   lastSelectedPieceName: string;
 
-  currentTypeOffs = 0;
+  currentTypeOffset = 0;
 
   segmentTypes = ['SEGMENT', 'PIECE', 'SUPER'];
   selectedSegmentType: PieceType;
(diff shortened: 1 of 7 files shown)


frontend/rolecall/src/app/piece/piece_editor.component.ts
```diff
@@ -197,7 +197,7 @@ export class PieceEditor implements OnInit {
     }
     this.currentSelectedPiece = piece;
     this.urlPointingUUID = piece ? piece.uuid : '';
-    this.currentType = this.getCurrentTypeCode(piece.type);
+    this.currentType = piece ? this.getCurrentTypeCode(piece.type) : 0;
```
mikzat, line 200: Does `0` have some meaning? Can we create a named constant to represent it to avoid the "magic number"?

mikzat (approved):

frontend/rolecall/src/app/piece/piece_editor.component.ts
yhedholm: It is the offset into a type name array for display purposes. 0 is an empty string.

yhedholm: I changed the name of the currentType field to to currentTypeOffset

PR #33, opened 2020-10-09
possible fix for customer error

yhedholm:
I found an instance where the system lost track of a newly created cast.

The fix involves
1) some renaming so it was clearer what was stored in an array,
2) fixing of some errors introduced by me earlier when we removed the deprecated null-and-undefined check,
3) reordering of some improperly ordered statements when a new cast is created.

I hope this fix will fix the customer error we discovered today.

### DIFF
--- a/frontend/rolecall/src/app/cast/cast-editor-v2.component.ts
+++ b/frontend/rolecall/src/app/cast/cast-editor-v2.component.ts
@@ -203,10 +203,10 @@ export class CastEditorV2 implements OnInit {
       index = this.selectedPieceCasts.findIndex(
           findCast => cast.uuid === findCast.uuid);
     }
-    if (index === undefined || index === -1) {
-      this.lastSelectedCastIndex = undefined;
-    } else {
+    if (index >= 0) {
       this.lastSelectedCastIndex = index;
+    } else {
+      this.lastSelectedCastIndex = undefined;
     }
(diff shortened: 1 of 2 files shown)


frontend/rolecall/src/app/cast/cast-editor-v2.component.ts
```diff
@@ -197,12 +197,13 @@ export class CastEditorV2 implements OnInit {
...
+    index?: number,
   }) {
-    if (!index && cast) {
-      index = this.filteredCasts.findIndex((cast) => cast.uuid === cast.uuid);
+    if (index === undefined && cast) {
+      index = this.selectedPieceCasts.findIndex(
+          findCast => cast.uuid === findCast.uuid);
     }
-    if (!index || index === -1) {
+    if (index === undefined || index === -1) {
```
mikzat, line 206: Can flip this as:
```
if (index >= 0) {
  this.lastSelectedCastIndex = index;
} else {
  this.lastSelectedCastIndex = undefined;
}
```

mikzat (approved):

frontend/rolecall/src/app/cast/cast-editor-v2.component.ts
yhedholm: Done.

PR #37, opened 2020-10-09
probably fixed customer bug

yhedholm:
I managed to reproduce the customer error on my local machine!

When I first implemented middle name and suffix, they couldn't be null. Later we changed that without my updating all places where we assumed middle name and suffix were not null. (It didn't help finding the bug that all "auto generated" names had middle names and suffixes that were non-null. We only ran into problems with the users we had to add manually because they had real emails.)

The good news is that I found many subtle error conditions that hopefully have been prevented by code fixes.

### DIFF
--- a/frontend/rolecall/src/app/cast/cast-drag-and-drop.component.ts
+++ b/frontend/rolecall/src/app/cast/cast-drag-and-drop.component.ts
@@ -141,11 +141,13 @@ export class CastDragAndDrop implements OnInit {
     uuid: APITypes.CastUUID;
     saveDeleteEnabled?: boolean;
   }) {
-    this.buttonsEnabled = saveDeleteEnabled;
-    this.castSelected = true;
-    this.selectedCastUUID = uuid;
-    if (this.dataLoaded) {
-      this.setupData();
+    if (uuid) {
+      this.buttonsEnabled = saveDeleteEnabled;
+      this.castSelected = true;
+      this.selectedCastUUID = uuid;
+      if (this.dataLoaded) {
+        this.setupData();
+      }
     }
   }
 
@@ -194,8 +196,8 @@ export class CastDragAndDrop implements OnInit {
     }
     this.castPositions = [];
     if (!this.castAPI.hasCast(this.selectedCastUUID)) {
-      this.castSelected = false;
       this.logging.logError("Missing cast: " + this.selectedCastUUID);
+      this.castSelected = false;
       return;
     }
     this.cast = this.castAPI.castFromUUID(this.selectedCastUUID);
--- a/frontend/rolecall/src/app/performance/performance-editor.component.ts
+++ b/frontend/rolecall/src/app/performance/performance-editor.component.ts
@@ -152,10 +152,12 @@ export class PerformanceEditor implements OnInit, OnDestroy, AfterViewChecked {
     this.initStep3Data();
     this.initStep4();
     this.state.status = PerformanceStatus.DRAFT;
-    // Saving casts results in an immediate cast load.
-    // Clear a key parameter to prevent deadly embrace
-    this.castDnD.castSelected = false;
-
+    if (this.castDnD) {
+      // Saving casts results in an immediate cast load.
+      // Clear key parameters prevent deadly embrace.
+      this.castDnD.castSelected = false;
+      this.castDnD.selectedCastUUID = undefined;
+    }
     this.performanceAPI.setPerformance(this.dataToPerformance());
     this.deleteWorkingCasts();
     this.resetState();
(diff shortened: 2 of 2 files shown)


frontend/rolecall/src/app/performance/performance-editor.component.ts
```diff
@@ -152,10 +152,12 @@ export class PerformanceEditor implements OnInit, OnDestroy, AfterViewChecked {
     this.initStep3Data();
     this.initStep4();
     this.state.status = PerformanceStatus.DRAFT;
-    // Saving casts results in an immediate cast load.
-    // Clear a key parameter to prevent deadly embrace
-    this.castDnD.castSelected = false;
-
+    if (this.castDnD) {
```
mikzat, line 155: Oof... interacting directly with ViewChild properties and methods... that's a pretty nasty design pattern. Not sure we'll be able to untangle this.

frontend/rolecall/src/app/performance/performance-editor.component.ts
```diff
@@ -152,10 +152,12 @@ export class PerformanceEditor implements OnInit, OnDestroy, AfterViewChecked {
...
+      // Saving casts results in an immediate cast load.
+      // Clear key parameters prevent deadly embrace.
+      this.castDnD.castSelected = false;
+      this.castDnD.selectedCastUUID = undefined;
```
mikzat, line 159: The `selectedCastUUID ` variable definition has the wrong type if it's valid to assign 'undefined' to it. If it's really valid, then the type must be changed to:
`selectedCastUUID: APITypes.CastUUID | undefined;`

frontend/rolecall/src/app/cast/cast-drag-and-drop.component.ts
```diff
@@ -141,11 +141,13 @@ export class CastDragAndDrop implements OnInit {
     uuid: APITypes.CastUUID;
     saveDeleteEnabled?: boolean;
   }) {
-    this.buttonsEnabled = saveDeleteEnabled;
-    this.castSelected = true;
-    this.selectedCastUUID = uuid;
-    if (this.dataLoaded) {
-      this.setupData();
+    if (uuid) {
```
mikzat, line 144: We should never pass undefined in to this function. It would be better if the callers did the check and never called this function at all if uuid is undefined. 

But if it's really unavoidable then the input type must be changed to:
`uuid: APITypes.CastUUID | undefined;`

mikzat (changes requested):

frontend/rolecall/src/app/performance/performance-editor.component.ts
yhedholm: The problem is that this routine, onSaveDraft(), in the next statement, removes the uuid that this.castDnd.selectedCastUUID is pointing at. (I don't want to change this.castDnD, I only want to undo the damage onSaveDraft is already doing). Basically when cast-drag-and-drop get control back after the save, it is going looking for selectedCastUUID and misbehaves when it can't find it.

I agree that it is a dirty coupling. The better way to solve this would probably be to rewrite the cast saving routine so that it somehow clears selectedCastUUID in the cast_api routines (that should be closer to cast-drag-and-drop than the performance-editor) but I don't understand the code on the right level to do that now.

frontend/rolecall/src/app/performance/performance-editor.component.ts
yhedholm: selectedCastUUID is already assigned undefined (in the delete routine and in several routines where a passed in function parameter can be undefined). An alternative would be to use and empty string as the "null value".

frontend/rolecall/src/app/cast/cast-drag-and-drop.component.ts
yhedholm: I copied in a routine that calls selectCast. It is in a different component so it has less information about the inner workings of selectCast. As you can see, it seems to be standard procedure to set variables that has no meaningful value to 'undefined'. Still, sometimes '' is used (for this.urlUUID).

It is quite possible to make sure that nobody calls selectCast with a null uuid. However, I feel that it probably makes sense to put in a test inside selectCast just to be sure.

  setCurrentCast({cast, index}: {
    cast: Cast | undefined,
    index?: number,
  }) {
    if (index === undefined && cast) {
      index = this.selectedPieceCasts.findIndex(
          findCast => cast.uuid === findCast.uuid);
    }
    if (index >= 0) {
      this.lastSelectedCastIndex = index;
    } else {
      this.lastSelectedCastIndex = undefined;
    }
    this.lastSelectedCast = cast ? cast : this.lastSelectedCast;
    this.selectedCast = cast;
    this.dragAndDrop.selectCast({uuid: cast ? cast.uuid : undefined});
    this.urlUUID = cast ? cast.uuid : '';
    this.setCastURL();
  }

yhedholm: When I was debugging the customer bug, selectCast was repeatedly called with an undefined uuid. I would like to put in a safeguard against bad things happening if selectCast is called with an undefined uuid.

I can create a new PR with only this change.

mikzat: Sure let's do a new clean PR with just that change and any related type changes if needed.

yhedholm:
Replaced by PR# 54

PR #45, opened 2020-10-12
Linted performance-editor.component.ts

yhedholm:
Everything was straightforward except "object access via string literals is disallowed" starting at line 319.

The automatic lint fix just converted the "['fieldName']" to ".fieldName". Because these fields aren't known to typescript, I had to cast to any to get it to build.

### DIFF
--- a/frontend/rolecall/src/app/performance/performance-editor.component.ts
+++ b/frontend/rolecall/src/app/performance/performance-editor.component.ts
@@ -1,17 +1,18 @@
-import { CdkDragDrop, copyArrayItem, transferArrayItem } from '@angular/cdk/drag-drop';
-import { Location } from '@angular/common';
-import { AfterViewChecked, ChangeDetectorRef, Component, OnDestroy, OnInit, ViewChild } from '@angular/core';
-import { MatSelectChange } from '@angular/material/select';
-import { ActivatedRoute } from '@angular/router';
-import { PerformanceStatus } from 'src/api_types';
-import { Cast, CastApi } from '../api/cast_api.service';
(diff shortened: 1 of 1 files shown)


RichardYang2016 (approved):

frontend/rolecall/src/app/performance/performance-editor.component.ts
```diff
@@ -297,53 +351,36 @@ export class PerformanceEditor implements OnInit, OnDestroy, AfterViewChecked {
...
-    else if (field == "venue") {
-      let val2 = value as InputEvent;
-      this.state.step_1.venue = val2.target['value'];
-    }
-    else if (field == "date") {
-      let val3 = value as InputEvent;
-      this.state.step_1.date = Date.parse(val3.target['value'])
+    const val = value as InputEvent;
+    if (field === 'title') {
+      this.state.step_1.title = (val.target as any).value;
```
mikzat, line 370: Could we do this just once before the if/else chain:
const targetValue = (val.target as any).value;

mikzat, line 370: The actual type of "val.target" looks like is "EventTarget | null", so let's completely avoid adding the `any` type as it could hide bugs. For example, how does it impact our code if the `target` property is actually null?

mikzat (changes requested):

frontend/rolecall/src/app/performance/performance-editor.component.ts
mikzat: Consider this comment optional, since the current PR was not intended to do any refactoring, just lints.

yhedholm:
Replaced by PR #47

PR #47, opened 2020-10-13
2nd part of linting performance-editor.component.ts

yhedholm:
From the earlier PR:

"Everything was straightforward except "object access via string literals is disallowed" starting at line 319. (in this PR, that became line 357.)

The automatic lint fix just converted the "['fieldName']" to ".fieldName". Because these fields aren't known to typescript, I had to cast to any to get it to build."

Because there were many lines that were over 80 characters wide, the diff reports lots of lines, but the only logic modifications are in onStep1Input starting at line 366 in the new file.

### DIFF
--- a/frontend/rolecall/src/app/performance/performance-editor.component.ts
+++ b/frontend/rolecall/src/app/performance/performance-editor.component.ts
@@ -1,17 +1,18 @@
-import { CdkDragDrop, copyArrayItem, transferArrayItem } from '@angular/cdk/drag-drop';
-import { Location } from '@angular/common';
-import { AfterViewChecked, ChangeDetectorRef, Component, OnDestroy, OnInit, ViewChild } from '@angular/core';
-import { MatSelectChange } from '@angular/material/select';
-import { ActivatedRoute } from '@angular/router';
-import { PerformanceStatus } from 'src/api_types';
-import { Cast, CastApi } from '../api/cast_api.service';
-import { Performance, PerformanceApi } from '../api/performance-api.service';
-import { Piece, PieceApi } from '../api/piece_api.service';
-import { User, UserApi } from '../api/user_api.service';
-import { CastDragAndDrop } from '../cast/cast-drag-and-drop.component';
-import { Stepper } from '../common_components/stepper.component';
-import { CsvGenerator } from '../services/csv-generator.service';
-import { ResponseStatusHandlerService } from '../services/response-status-handler.service';
+import {CdkDragDrop, copyArrayItem, transferArrayItem} from '@angular/cdk/drag-drop';
+import {Location} from '@angular/common';
+import {AfterViewChecked, ChangeDetectorRef, Component, OnDestroy,
+    OnInit, ViewChild} from '@angular/core';
+import {MatSelectChange} from '@angular/material/select';
+import {ActivatedRoute} from '@angular/router';
+import {PerformanceStatus} from 'src/api_types';
+import {Cast, CastApi} from '../api/cast_api.service';
+import {Performance, PerformanceApi} from '../api/performance-api.service';
+import {Piece, PieceApi} from '../api/piece_api.service';
+import {User, UserApi} from '../api/user_api.service';
+import {CastDragAndDrop} from '../cast/cast-drag-and-drop.component';
+import {Stepper} from '../common_components/stepper.component';
+import {CsvGenerator} from '../services/csv-generator.service';
+import {ResponseStatusHandlerService} from '../services/response-status-handler.service';
 import {CAST_COUNT} from 'src/constants';
 
 @Component({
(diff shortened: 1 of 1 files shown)


frontend/rolecall/src/app/performance/performance-editor.component.ts
```diff
@@ -21,8 +22,18 @@ import {CAST_COUNT} from 'src/constants';
 })
 export class PerformanceEditor implements OnInit, OnDestroy, AfterViewChecked {
 
+  constructor(private performanceAPI: PerformanceApi,
```
mikzat, line 25: Let's move this back down as the first function in the file, so it's not mixed in with the properties

frontend/rolecall/src/app/performance/performance-editor.component.ts
```diff
@@ -346,42 +356,30 @@ export class PerformanceEditor implements OnInit, OnDestroy, AfterViewChecked {
...
-    else if (field == "venue") {
-      let val2 = value as InputEvent;
-      this.state.step_1.venue = val2.target['value'];
-    }
-    else if (field == "date") {
-      let val3 = value as InputEvent;
-      this.state.step_1.date = Date.parse(val3.target['value'])
+    const val = value as InputEvent;
+    if (field === 'title') {
+      this.state.step_1.title = (val.target as any).value;
```
mikzat, line 369: Same comment as on #45 , "The actual type of "val.target" looks like is "EventTarget | null", so let's completely avoid adding the any type as it could hide bugs. For example, how does it impact our code if the target property is actually null?"

frontend/rolecall/src/app/performance/performance-editor.component.ts
```diff
@@ -532,28 +545,34 @@ export class PerformanceEditor implements OnInit, OnDestroy, AfterViewChecked {
...
     this.primaryGroupNum = castAndPrimLength[1];
-    this.castDnD ? this.castDnD.selectCast({uuid: castUUID, saveDeleteEnabled: false}) : '';
+    if (this.castDnD) {
+      this.castDnD.selectCast({uuid: castUUID, saveDeleteEnabled: false});
+    }
     this.updateGroupIndices(castAndPrimLength[0]);
     this.updateCastsForSegment();
     this.segmentLength = castAndPrimLength[2];
     this.changeDetectorRef.detectChanges();
-    this.castDnD ? this.castDnD.setBoldedCast(this.segmentToCast.get(castUUID)[1]) : '';
```
mikzat, line 540: What a strange syntax this was, wonder why they had it this way...

frontend/rolecall/src/app/performance/performance-editor.component.ts
```diff
@@ -532,28 +545,34 @@ export class PerformanceEditor implements OnInit, OnDestroy, AfterViewChecked {
       castAndPrimLength = this.segmentToCast.get(castUUID);
     }
     this.primaryGroupNum = castAndPrimLength[1];
-    this.castDnD ? this.castDnD.selectCast({uuid: castUUID, saveDeleteEnabled: false}) : '';
+    if (this.castDnD) {
+      this.castDnD.selectCast({uuid: castUUID, saveDeleteEnabled: false});
```
mikzat, line 549: A better way to write this would be to get rid of the `if (this.castDnD)` and just have:
`this.castDnD?.selectCast({uuid: castUUID, saveDeleteEnabled: false});`

mikzat (changes requested):

frontend/rolecall/src/app/performance/performance-editor.component.ts
mikzat: By the way, this applies to several sections in the current PR

frontend/rolecall/src/app/performance/performance-editor.component.ts
yhedholm: Done.

frontend/rolecall/src/app/performance/performance-editor.component.ts
yhedholm: I believe the value is always an input event and never null. I tried to define it as an HTMLInputEvent but the typescript can't find HTMLInputEvent.

So I implemented your original suggestion.

frontend/rolecall/src/app/performance/performance-editor.component.ts
yhedholm: I believe they were trying to do what your "better way" syntax below did.

frontend/rolecall/src/app/performance/performance-editor.component.ts
yhedholm: Done.

mikzat (approved):

PR #50, opened 2020-10-13
Enable strictFunctionTypes in tsconfig and fix related failure.

mikzat:


frontend/rolecall/src/app/piece/piece_editor.component.ts
```diff
@@ -152,12 +152,15 @@ export class PieceEditor implements OnInit {
...
-      val.addingPositions = [];
-      val.originalName = String(val.name);
-      return val;
-    });
+
+    const workPieces = pieces.map(piece => ({
+      ...piece,
+      addingPositions: [],
+      originalName: String(piece.name),
+      isOpen: false,
```
mikzat, line 160: Need confirmation if this is the correct default value.

yhedholm: Pieces come in three kinds: Segment, Ballet, and Super Ballet. Only Super Ballets pay attention to the isOpen field. (Only Super Ballets can "open up and show their children". So the default value for Segments and Ballets really doesn't matter but false is fine.) For Super Ballets, the default value when first created should be 'true' so that people can see the Ballet children.

mikzat: As discussed offline, for now will default `isOpen` to false and if needed logic will be revisited in the future.

PR #51, opened 2020-10-14
Lint fixes in services

mikzat:


frontend/rolecall/src/app/api/cast_api.service.ts
```diff
@@ -82,278 +84,318 @@ export type OneCastResponse = {
...
+                          return {
+                            uuid: String(rawMem.userId),
+                            position_number: rawMem.order
+                          };
+                        })
+                      });
+                    }
                   } else {
                     groups.push({
-                      position_uuid: String(allPositions.find(pos => Number(pos.uuid) == rawSubCast.positionId).uuid),
```
yhedholm, line 152: It seems as if the lines at 188 through 201 are the same as 174 through 187.

yhedholm (changes requested):
Thanks for doing this! I believe I caught some code duplication in cast_api.service.ts

frontend/rolecall/src/app/api/cast_api.service.ts
mikzat: Good catch, I added a TODO above to resolve this later, don't want to mess with logic in the same PR

RichardYang2016:
not familiar yet with logic but just clear for lint fixes.

RichardYang2016 (approved):

yhedholm (approved):

PR #53, opened 2020-10-14
Enable noUnusedLocals and fix related issue

mikzat:
Minor cleanup in cast drag+drop component.


yhedholm (approved):
I believe I already approved these (or something very similar) as part of PR# 52.

PR #54, opened 2020-10-14
Some leftover fixes

yhedholm:
Some leftover fixes:
1) Remove some users from our temporary pre-populate list.
2) Secure selectCast so that calls with uuid being undefined don't wreak havoc with the system.
3) Some simple reformatting to match Mikhail's constructor formatting.
4) Removal of a cast to any.

### DIFF
--- a/frontend/rolecall/src/app/performance/performance-editor.component.ts
+++ b/frontend/rolecall/src/app/performance/performance-editor.component.ts
@@ -85,15 +85,17 @@ export class PerformanceEditor implements OnInit, OnDestroy, AfterViewChecked {
 
   // --------------------------------------------------------------
 
-  constructor(private performanceAPI: PerformanceApi,
-              private piecesAPI: PieceApi,
-              private castAPI: CastApi,
-              private respHandler: ResponseStatusHandlerService,
-              private userAPI: UserApi,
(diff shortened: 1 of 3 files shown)


frontend/rolecall/src/app/performance/performance-editor.component.ts
```diff
@@ -364,7 +366,7 @@ export class PerformanceEditor implements OnInit, OnDestroy, AfterViewChecked {
   }
 
   onStep1Input(field: string, event: InputEvent) {
-    const targetValue = (event.target as any).value;
+    const targetValue = (event.target as HTMLInputElement).value;
```
mikzat, line 369: Just a note here that `event.target` might be null according to the type: `readonly target: EventTarget | null;`, so by forcing it into `HTMLInputElement`  there's theoretically a chance we'll hit a null value in this function at runtime.

Maybe just to be safe we can also add
```
if (!event.target) {
  return;
}
```

mikzat (approved):

frontend/rolecall/src/app/performance/performance-editor.component.ts
yhedholm: OK.

PR #57, opened 2020-10-14
Frontend auto-format changes only

mikzat:


yhedholm (approved):
It seems that this reformatting run formatted constructors differently than you normally do. You normally do:
constructor(
    xxx,
    yyy);
This time it did:
constructor(xxx,
                     yyy);

PR #68, opened 2020-10-14
Remove namespace usage with minimal impact on existing code

mikzat:
Would love to get rid of the APITypes prefix completely, but it can be considered as a separate PR in the future.


yhedholm (approved):
Thanks!

PR #71, opened 2020-10-15
Lint fixes for login_api.service.spec

yhedholm:
As part of this, I moved some type files from dependencies to devDependencies in package.json.


frontend/rolecall/src/app/api/login_api.service.spec.ts
```diff
@@ -80,8 +79,8 @@ describe('LoginApi', () => {
     await loginRespProm;
     mockObj.isSignedInVal = false;
 
-    expect(isNullOrUndefined(service.authInstance.currentUser.get()))
-        .toBeTrue();
+    expect(service?.authInstance?.currentUser?.get())
+        ?.toBeTrue();
```
mikzat, line 83: I think this is supposed to check that the result is null, so possibly you want something like this:

```
expect(service.authInstance.currentUser.get()).toBeNull();
```

If the intermediate steps are undefined, then the test will also fail correctly.

mikzat (changes requested):

frontend/rolecall/src/app/api/login_api.service.spec.ts
yhedholm: Changed.

mikzat (approved):

PR #75, opened 2020-10-17
Add copy database sql code

yhedholm:
This code copies all data and compresses the index values by resetting the AUTO_INCREMENT counter for every table. By editing this code, we can support making structural changes to the database without losing customer data.


backend/src/main/resources/convert_database.txt
```diff
@@ -0,0 +1,568 @@
+/* SQL make a copy of the rolcall_db database in the database rolecall_new. */
```
kippig, line 1: Do we want this to be a .sql file?

kippig (approved):

PR #76, opened 2020-10-17
Made SuperBallet display state persistent

yhedholm:
The changes we did to the SuperBallet display did change the open behavior. Every time a ballet was saved, the SuperBallets' open state would be set to false. (I had hoped this would happen much more infrequently.)

This fix involves creating a service where the SuperBallets' open states are remembered throughout a session.


frontend/rolecall/src/app/piece/piece_editor.component.ts
```diff
@@ -99,7 +100,9 @@ export class PieceEditor implements OnInit {
     this.displayedPieces = this.workingPieces.filter(piece => !piece.siblingId);
     this.displayedPieces.sort((a, b) => a.name < b.name ? -1 : 1);
     for (let i = 0; i < this.displayedPieces.length; i++) {
-      if (this.displayedPieces[i].isOpen) {
+      if (this.displayedPieces[i].type === 'SUPER' &&
```
mikzat, line 103: Referencing `this.displayedPieces[i]` four times, it'll be a little more readable if you first set
```
const piece = this.displayedPieces[i];
```
and use that in the four places

frontend/rolecall/src/app/services/super-display.service.ts
```diff
@@ -0,0 +1,30 @@
...
+  superIsOpenMap = new Map<string, boolean>();
+
+  delete(uuid: string) {
+    if (this.superIsOpenMap.has(uuid)) {    
+      this.superIsOpenMap.delete(uuid);
+    }    
+  }
+
+  isOpen(uuid: string): boolean {
+    if (!this.superIsOpenMap.has(uuid)) {
```
mikzat, line 15: You can optionally simplify this to:
```
isSuperBalletOpen(balletId: string): boolean {
  return Boolean(this.superBalletIsOpenMap.get(balletId);
}
```
It'll handle both missing and false values. But you can leave the conditions separated if you want to be more explicit.

frontend/rolecall/src/app/piece/piece_editor.component.ts
```diff
@@ -157,12 +160,17 @@ export class PieceEditor implements OnInit {
...
 
-    const workPieces = pieces.map(piece => ({
-      ...piece,
-      addingPositions: [],
-      originalName: String(piece.name),
-      isOpen: false,
-    }));
+    const workPieces = pieces.map(piece => {
+      if (piece.type === 'SUPER') {
+        this.superDisplay.verifyLoad(piece.uuid, false);
```
mikzat, line 165: This should be something like
```
this.balletDisplayService.closeSuperBallet(piece.uuid);
```

frontend/rolecall/src/app/piece/piece_editor.component.ts
```diff
@@ -79,7 +79,8 @@ export class PieceEditor implements OnInit {
       private route: ActivatedRoute,
       private pieceAPI: PieceApi,
       private location: Location,
-      private respHandler: ResponseStatusHandlerService) {
+      private respHandler: ResponseStatusHandlerService,
+      private superDisplay: DisplayService) {
```
mikzat, line 83: This should be more specific like superBalletDisplayService or balletDisplayService

frontend/rolecall/src/app/piece/piece_editor.component.ts
```diff
@@ -273,6 +281,7 @@ export class PieceEditor implements OnInit {
       addingPositions: [],
       deletePositions: [],
     };
+    this.superDisplay.verifyLoad(newPiece.uuid, newPiece.isOpen);
```
mikzat, line 284: This would be more readable: 
```
if (newPiece.type === 'SUPER') {
  this.balletDisplayService.openSuperBallet(newPiece.uuid);
}
```

mikzat (changes requested):

frontend/rolecall/src/app/performance/performance-editor.component.ts
```diff
@@ -120,7 +120,6 @@ export class PerformanceEditor implements OnInit, OnDestroy, AfterViewChecked {
   }
 
   closeStepper() {
```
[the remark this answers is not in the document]
yhedholm: Done. I checked for callers. There were none. (There was another function with exactly the same name called from HTML code. I believe the printout was there to verify that this routine was never called.)

frontend/rolecall/src/app/services/super-display.service.spec.ts
```diff
@@ -0,0 +1,16 @@
+import {TestBed} from '@angular/core/testing';
+
+import {DisplayService} from './super-display.service';
+
+describe('DisplayService', () => {
+  let service: DisplayService;
+
+  beforeEach(() => {
+    TestBed.configureTestingModule({});
```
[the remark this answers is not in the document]
yhedholm: Done.

frontend/rolecall/src/app/services/super-display.service.ts
```diff
@@ -0,0 +1,30 @@
+import {Injectable} from '@angular/core';
+
+
+@Injectable({providedIn: 'root'})
+export class DisplayService {
```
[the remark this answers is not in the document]
yhedholm: Done.

frontend/rolecall/src/app/services/super-display.service.ts
```diff
@@ -0,0 +1,30 @@
...
+  superIsOpenMap = new Map<string, boolean>();
```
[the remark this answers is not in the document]
yhedholm: Done.

frontend/rolecall/src/app/services/super-display.service.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  delete(uuid: string) {
```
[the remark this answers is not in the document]
yhedholm: I made new names for the functions:
changeOpenState(uuid: string, isOpen: boolean) // changes open state
isOpen(uuid: string): boolean // returns current open state
removeFromDisplayList(uuid: string) // removes super ballet from display list
verifyIsInDisplayList(uuid: string, isOpen: boolean) // adds a super ballet to display list if not already there

frontend/rolecall/src/app/piece/piece_editor.component.ts
yhedholm: Done.

frontend/rolecall/src/app/services/super-display.service.ts
yhedholm: Done.

frontend/rolecall/src/app/piece/piece_editor.component.ts
yhedholm: Done.

frontend/rolecall/src/app/piece/piece_editor.component.ts
yhedholm: With the new naming this is now:

this.superBalletDisplay.verifyIsInDisplayList(newPiece.uuid, newPiece.isOpen);

It is important that already displayed super ballet don't change their open state when they are being re-read from the database, which happens after each save of a new or edited ballet. Because of this, the logic of the verifyIsInDisplayList is that ballets already in the display list are left untouched. However, ballets not in the display list are added in two ways. Ballets from the database go in as closed. Newly created ballets go in as open. (The customers can change their name about this, but at this point they have asked for the open state.)

frontend/rolecall/src/app/piece/piece_editor.component.ts
yhedholm: With the new name this now looks like

this.superBalletDisplay.verifyIsInDisplayList(piece.uuid, false);

The logic is, if the super ballet is not already in the display list, it will be added in the closed state. If is is already in the display list, its open state will not be altered.

frontend/rolecall/src/app/services/super-ballet-display.service.spec.ts
```diff
@@ -0,0 +1,13 @@
+import {SuperBalletDisplayService} from './super-ballet-display.service';
+
+describe('SuperBalletDisplayService', () => {
+  let service: SuperBalletDisplayService;
+
+  beforeEach(() => {
+    service = new SuperBalletDisplayService;
```
mikzat, line 7: If you run `ng test` you will get a failure with this file due to missing parenthesis. Always make sure your tests pass :)

frontend/rolecall/src/app/services/super-ballet-display.service.ts
```diff
@@ -0,0 +1,27 @@
...
+  changeOpenState(uuid: string, isOpen: boolean) {
+    this.superBalletIsOpenMap.set(uuid, isOpen);
+  }
+
+  isOpen(uuid: string): boolean {
+    return Boolean(this.superBalletIsOpenMap.get(uuid));
+  }
+
+  removeFromDisplayList(uuid: string) {
+    if (this.superBalletIsOpenMap.has(uuid)) {
```
mikzat, line 17: I don't think you need the if() check here, it can probably be just
```
removeFromDisplayList(balletId: string) {
  this.superBalletIsOpenMap.delete(balletId);
}
```

According to [the docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/delete), delete() will just return false if the prop didn't exist.

frontend/rolecall/src/app/services/super-ballet-display.service.ts
```diff
@@ -0,0 +1,27 @@
+import {Injectable} from '@angular/core';
+
+
+@Injectable({providedIn: 'root'})
+export class SuperBalletDisplayService {
+  superBalletIsOpenMap = new Map<string, boolean>();
+
+  changeOpenState(uuid: string, isOpen: boolean) {
```
mikzat, line 8: What do you think about calling it "setOpenState"? It's a tiny bit shorter and a little closer aligned with what it's doing.

mikzat, line 8: I would prefer to avoid using the word `uuid` in the UI, calling it `balletId` is better semantically as it gives more context to what it's identifying.

frontend/rolecall/src/app/services/super-ballet-display.service.ts
```diff
@@ -0,0 +1,27 @@
...
+    return Boolean(this.superBalletIsOpenMap.get(uuid));
+  }
+
+  removeFromDisplayList(uuid: string) {
+    if (this.superBalletIsOpenMap.has(uuid)) {    
+      this.superBalletIsOpenMap.delete(uuid);
+    }    
+  }
+
+  verifyIsInDisplayList(uuid: string, isOpen: boolean) {
```
mikzat, line 22: Need to discuss this one, naming isn't right and I'm not clear about purpose. Something with a prefix of "verify" should not have any state side-effects, I would expect it to either throw an exception if verification fails, or possibly return a boolean.

mikzat (changes requested):

frontend/rolecall/src/app/services/super-ballet-display.service.ts
yhedholm: Done.

frontend/rolecall/src/app/services/super-ballet-display.service.spec.ts
yhedholm: Thanks. Fixed.

frontend/rolecall/src/app/services/super-ballet-display.service.ts
yhedholm: Done.

frontend/rolecall/src/app/services/super-ballet-display.service.ts
yhedholm: Done.

mikzat (approved):

PR #77, opened 2020-10-17
Change delete logic for Super Ballets

yhedholm:
One request we received from the customer on Thursday was that we change the delete logic for Super Ballets. Originally, we weren't sure how closely the customer wanted to couple Super Ballets and their children. If a Super Ballet is killed, should it be possible for its children to survive as independent entities? Originally, we thought the answer was yes.

The customer told us on Thursday that the answer is no. If you kill a Super Ballet, they expected all children to be killed as well.

This PR fixes the delete issue. I also fixed an assignment mistake where the "position size" (an important value for a position but a nonsense value for a super ballet child) was set equal to the siblingId -- probably cased by a sloppy copy paste. I changed it to null, which makes more sense.

I apologize for giving you backend code to review, but I really like your reviews. Besides, you reviewed this code before, so you probably half know it already.

### DIFF
--- a/backend/src/main/java/com/google/rolecall/services/SectionServices.java
+++ b/backend/src/main/java/com/google/rolecall/services/SectionServices.java
@@ -279,10 +279,40 @@ public class SectionServices {
    *    in the database.
    */
   public void deleteSection(int id) throws EntityNotFoundException, InvalidParameterException {
+    checkDeleteOk(id, false);
+    Section section = getSection(id);
+
+    if(section.getType() == Section.Type.SUPER) {
+      // Check that all children are OK to delete
+      for(Position position: section.getPositions()) {
+        Integer childId = position.getSiblingId();
+        if(childId != 0 && childId > 0) {
+          Optional<Section> childQuery = sectionRepo.findById(childId);
+          if(!childQuery.isEmpty()) {
+            checkDeleteOk(childId, true);
+          }
+        }
+      }
+      // If none of the children have a deleate problem, we can delete
+      for(Position position: section.getPositions()) {
+        Integer childId = position.getSiblingId();
+        if(childId != 0 && childId > 0) {
+          Optional<Section> childQuery = sectionRepo.findById(childId);
+          if(!childQuery.isEmpty()) { 
+            sectionRepo.deleteById(childId);
+          }
+        }
+      }
+    }  
+    sectionRepo.deleteById(id);
+  }
+
+  // Utility functions
+  private void checkDeleteOk(int id, boolean superChildIsOk) throws EntityNotFoundException, InvalidParameterException {
     Section section = getSection(id);
 
     Integer siblingId = section.getSiblingId();
-    if(siblingId != null) {
+    if(!superChildIsOk && siblingId != null) {
       Optional<Position> test = positionRepo.findById(siblingId);
       if(test.isEmpty()) {
         // Super Ballet's internal Ballet/Position structure has already been deleted
@@ -298,11 +328,7 @@ public class SectionServices {
       throw new InvalidParameterException(
           "Cannot delete Ballet if it has a Cast or is part of a Performance");
     }
-
-    sectionRepo.deleteById(id);
   }
-
-  // Utility functions
 
   private void updateSuperBalletChildren(Section section, Integer[] siblingIndexArray,
       boolean isParentSuper) throws InvalidParameterException {
@@ -343,7 +369,7 @@ public class SectionServices {
               .setNotes(sibling.getNotes())
               .setOrder(sibling.getOrder())
               .setSiblingId(siblingId == -1 ? sibling.getSiblingId() : siblingId)
-              .setSize(sibling.getSiblingId())
+              .setSize(null)
               .build();
           positionRepo.save(updatedSibling);
         } catch (InvalidParameterException e) {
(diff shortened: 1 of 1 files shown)


backend/src/main/java/com/google/rolecall/services/SectionServices.java
```diff
@@ -343,7 +369,7 @@ private SiblingError updateSiblingPosition(Integer positionId,
               .setNotes(sibling.getNotes())
               .setOrder(sibling.getOrder())
               .setSiblingId(siblingId == -1 ? sibling.getSiblingId() : siblingId)
-              .setSize(sibling.getSiblingId())
+              .setSize(null)
```
mikzat, line 372: If it's possible for this to be null, then the frontend type needs to also be adjusted, right? Currently it looks like we have it defined as `size: number` which means it can never be null or undefined, and if it ever is then we could have bugs on the frontend.

backend/src/main/java/com/google/rolecall/services/SectionServices.java
```diff
@@ -279,10 +279,40 @@ public Section editSection(SectionInfo newSection) throws EntityNotFoundExceptio
...
+            sectionRepo.deleteById(childId);
+          }
+        }
+      }
+    }  
+    sectionRepo.deleteById(id);
+  }
+
+  // Utility functions
+  private void checkDeleteOk(int id, boolean superChildIsOk) throws EntityNotFoundException, InvalidParameterException {
```
mikzat, line 311: I feel like the `Ok` suffix in the function name and in the `superChildIsOk ` is ambiguous. Reading this function on its own, it could mean:
"has been deleted successfully"
"some object has proper structure"
"operation is ready to be performed"

I suggest changing it to something more meaningful, like `assertDeleteAllowed()` or maybe something else that makes more sense.

mikzat, line 311: Might be helpful to add a javadoc to this to describe what it's for and what the args are

backend/src/main/java/com/google/rolecall/services/SectionServices.java
```diff
@@ -279,10 +279,40 @@ public Section editSection(SectionInfo newSection) throws EntityNotFoundExceptio
...
    */
   public void deleteSection(int id) throws EntityNotFoundException, InvalidParameterException {
+    checkDeleteOk(id, false);
+    Section section = getSection(id);
+
+    if(section.getType() == Section.Type.SUPER) {
+      // Check that all children are OK to delete
+      for(Position position: section.getPositions()) {
+        Integer childId = position.getSiblingId();
+        if(childId != 0 && childId > 0) {
```
mikzat, line 289: Seems like just `if (childId > 0)` is sufficient here?

backend/src/main/java/com/google/rolecall/services/SectionServices.java
```diff
@@ -279,10 +279,40 @@ public Section editSection(SectionInfo newSection) throws EntityNotFoundExceptio
...
+          Optional<Section> childQuery = sectionRepo.findById(childId);
+          if(!childQuery.isEmpty()) {
+            checkDeleteOk(childId, true);
+          }
+        }
+      }
+      // If none of the children have a deleate problem, we can delete
+      for(Position position: section.getPositions()) {
+        Integer childId = position.getSiblingId();
+        if(childId != 0 && childId > 0) {
```
mikzat, line 299: Same

backend/src/main/java/com/google/rolecall/services/SectionServices.java
```diff
@@ -279,10 +279,40 @@ public Section editSection(SectionInfo newSection) throws EntityNotFoundExceptio
...
+          Optional<Section> childQuery = sectionRepo.findById(childId);
+          if(!childQuery.isEmpty()) {
+            checkDeleteOk(childId, true);
+          }
+        }
+      }
+      // If none of the children have a deleate problem, we can delete
+      for(Position position: section.getPositions()) {
```
mikzat, line 297: I think all these extra `findById` queries can be avoided if you keep a list of childIds, append to it right after `checkDeleteOk(childId, true);`, and here just loop through that list and simply call `sectionRepo.deleteById(childId);`

Seems like that would be better performance, unless I missed something.

mikzat (changes requested):

backend/src/main/java/com/google/rolecall/services/SectionServices.java
yhedholm: Good catch! I changed null to -1 which also matched what I did elsewhere.

backend/src/main/java/com/google/rolecall/services/SectionServices.java
yhedholm: Done.

yhedholm: Done.

backend/src/main/java/com/google/rolecall/services/SectionServices.java
yhedholm: Fixed. Thanks!

backend/src/main/java/com/google/rolecall/services/SectionServices.java
yhedholm: Fixed.

backend/src/main/java/com/google/rolecall/services/SectionServices.java
yhedholm: I believe what I just did solves the findbyId issue.

PR #78, opened 2020-10-18
Fixed several todos and a few extra errors

yhedholm:
The only interesting one was a color problem. In the big auto css cleanup, we combined border specifications to be on one line rather than several. When it came to the circles in the Performance Step Indicator, the separate line specification yielded gray circles when they should be deep orange. Still, that was better than the same line specification, which yielded no circles! The error was apparently with the color even though I don't know exactly why. It could be because $AAOrange is defined as a linear gradient. When I created a standard color $AABorderColor: #ff6666; everything worked.


frontend/rolecall/src/app/services/response-status-handler.service.ts
```diff
@@ -65,6 +65,7 @@ export class ResponseStatusHandlerService {
   showError(errorEvent: ErrorEvent): Promise<string> {
     if (this.pendingErrors.has(errorEvent.url)) {
       // TODO: What should we return here?
+      // Is here a problem?
```
mikzat, line 68: The problem here is that we are returning an empty string, which is definitely not the intention of a function with `Promise<string>` type. The questions here are why is it returning a string in the first place, what are we doing with that string, and maybe we should change the return type to just void? It's not obvious to me what the return value means here.

But this fix might be out of scope of the current PR, so many just remove the added comment to get the rest of the fixes in?

frontend/rolecall/src/app/login/welcome-page.component.html
```diff
@@ -15,8 +15,7 @@ <h1 style="color:#FFFFFF; font-size:800%; font-weight:200; font-family:Tahoma, G
           <!-- Divider -->
           <hr>
           <!-- Description -->
-          <!-- TODO: Fix invalid font-size pixel value. -->
-          <h2 style="color:#FFFFFF; font-size: 200 px; font-weight: 200; font-family:Tahoma, Geneva, sans-serif">
+          <h2 style="color:#FFFFFF; font-size: 200%; font-weight: 200; font-family:Tahoma, Geneva, sans-serif">
```
mikzat, line 18: Please double check that 200% was intended here instead of 200px.

frontend/rolecall/src/styles/_colors.scss
```diff
@@ -6,3 +6,4 @@ $gray: #6d6d6d;
 $lightGray: #9c9c9c;
 $lightLightGray: #b1b1b1;
 $black: black;
+$AAOrangeBorder: #ff6666;
```
mikzat, line 9: Let's call the variable `$AAOrangeColor` because I would expect something with a suffix of `Border` to be defining a border style like `1px solid #f3f3f3`, not just a color. Also let's move it up next to `$AAOrange` since they are closely related.

It would even be great to rename `$AAOrange` to `$AAOrangeGradient` to avoid the naming confusion in the future.

frontend/rolecall/src/app/user/user-editor.component.ts
```diff
@@ -71,8 +71,7 @@ export class UserEditor implements OnInit {
 
   rolesNamesMap = {
     isAdmin: 'Is Admin',
-    // TODO: Spelling error in `choreographer`.
-    isCoreographer: 'Is Coreographer',
+    isCoreographer: 'Is Choreographer',
```
mikzat, line 74: I imagine fixing the typo in the property name is considerably more invasive?

mikzat (changes requested):

frontend/rolecall/src/app/services/response-status-handler.service.ts
yhedholm: I thought the empty promise is like a promise no-op, or a promise when you know the returned value will be ignored. (Like the return value from a message dialog where OK is the only option.) This promise comes after displaying an error dialog and apparently the user has no option except to click OK.) I am happy to remove the TODO.

yhedholm: I removed the extra comment but left the todo.

frontend/rolecall/src/app/user/user-editor.component.ts
yhedholm: I missed that!

yhedholm: Actually, this is a biggie. I will make a special PR out of it.

frontend/rolecall/src/app/login/welcome-page.component.html
yhedholm: It was intended.

frontend/rolecall/src/styles/_colors.scss
yhedholm: I made the changes you suggested.

mikzat (approved):

PR #79, opened 2020-10-19
coreographer to choreographer

yhedholm:


backend/src/main/java/com/google/rolecall/Constants.java
```diff
@@ -32,11 +32,11 @@
 
   public static class Roles {
     public static final String ADMIN = "ADMIN";
-    public static final String COREOGRAPHER = "COREOGRAPHER";
```
mikzat, line 35: Is there any impact on the DB from this change?

frontend/rolecall/src/app/user/user-editor.component.ts
```diff
@@ -72,7 +72,7 @@ export class UserEditor implements OnInit {
   rolesNamesMap = {
     isAdmin: 'Is Admin',
     // TODO: Spelling error in `choreographer`.
```
mikzat, line 74: I believe the TODO will be removed from the other PR, but wouldn't hurt to remove it here as well.

mikzat (approved):

backend/src/main/java/com/google/rolecall/Constants.java
yhedholm: No. It seems as if the original developers changed strategy at some point and ignored their system wide constants. CHOREOGRAPHER only shows up in one place, inside a comment!

frontend/rolecall/src/app/user/user-editor.component.ts
yhedholm: Done.

PR #80, opened 2020-10-19
Added two fields to the User table

yhedholm:
I added notificationEmail and pictureFile to the User table.

The addition of the notificationEmail is preemptive. Database changes involve extra work, so the fewer occasions the better.

The pictureFile change was requested by the customer. Right now the picture file name is generated by the system as firstName_lastName. The problem with this is 1) we got some of the names wrong (some hyphened last names and non standard characters etc.), and 2) people change their names for many reasons. Moreover, artists are very picky with how their names are displayed, even internally (after all, it is their brand). Our new logic, to be implemented later on the front end, is to store the picture file name separately so the dancers' names can be edited freely.

I also changed the database name, so we can keep the current database around until we are sure our conversion program worked perfectly.

RIchard, I added you to the PR because Mikhail is extra busy right now. It is interesting how much code you have to add to add two fields / columns to a database table.


kippig (commented):
LGTM

kippig (approved):

PR #86, opened 2020-10-27
Picture File Name and Notification Email frontend

yhedholm:
Richard, I am making you the reviewer because Mikhail is busy until next Monday.

These are front end changes and a few trivial bug fixes on the backend associated with adding 1) a picture file name field, and 2) a notification email field to the user screen.

The largest # diff lines come from the update of mock data. Other than that, the changes are pretty straight forward.


RichardYang2016 (approved):
I feel code is ok. just a general question on CSS. are we not using any grid system?

PR #88, opened 2020-11-01
Organizing prod and cloud CICD environments

RichardYang2016:
Organizing prod and cloud CICD environments


yhedholm (approved):
Let's try the code and see if it works.

PR #89, opened 2020-11-01
fix index error

yhedholm:
After converting the database, we got index errors whenever we tried to save a cast. The system complained about an index mis-match.

In order to debug this, I decided to remove the SET FOREIGN_KEY_CHECKS=0 statement.  That, in turn forced me to re-order the table create statements so that the most primitive table is created first etc. (Before I ordered the creates in alphabetic order.) Eventually, I found a that the error was a mismatch in the CastMember table. The cast_id field is actually not a cast_id but a subCast_id. (I tried to changed the data column name to subCast_id, but I had to revert because I couldn't make hibernate happy.)

Anyway, the diff has lots of changes due to the reordering of table create statements. I also added some diagnostic test at the end.  Still, the only effective change is that the cast_id index is now pointing correctly to the SubCast table.


RichardYang2016 (approved):

backend/src/main/resources/convert_database.txt
```diff
@@ -566,3 +644,31 @@ DROP TABLE IF EXISTS Section_Id_Lookup;
 DROP TABLE IF EXISTS SubCast_Id_Lookup;
 DROP TABLE IF EXISTS Unavailability_Id_Lookup;
 DROP TABLE IF EXISTS User_Id_Lookup;
+DROP TABLE IF EXISTS MaxIds;
+
+CREATE DATABASE rolecall_old;
```
zephyr-l, line 649: I think the backup should go to the top of the script.

zephyr-l (approved):

PR #90, opened 2020-11-01
New full name pipe

yhedholm:
A while back you suggested we create a fullName pipe. At the time we were scrambling, so I decided to wait. I now implemented the fullName pipe.


backend/src/main/java/com/google/rolecall/services/CastServices.java
```diff
@@ -61,7 +61,9 @@ public Cast getCast(Integer id) throws EntityNotFoundException, InvalidParameter
     Cast cast = createNewCast(newCast);
     List<String> warnings = verifySubCasts(cast);
 
+    System.out.printf("AAA\n");
```
mikzat, line 64: let's remove these debugging print outs

frontend/rolecall/src/app/cast/cast-editor-v2.component.ts
```diff
@@ -81,7 +81,7 @@ export class CastEditorV2 implements OnInit {
     let autoSelectFirst = false;
     this.selectedPiece = piece;
     this.updateFilteredCasts();
-    if (this.selectedPiece && this.selectedPiece.uuid !== piece.uuid) {
```
mikzat, line 84: Is this change required for the new pipe implementation? If not, could it be moved to a separate PR?

frontend/rolecall/src/app/common_components/full-name.pipe.ts
```diff
@@ -0,0 +1,19 @@
+import {Pipe, PipeTransform} from '@angular/core';
+import {User} from '../api/user_api.service';
+
+@Pipe({
+  name: 'fullName'
+})
+export class FullNamePipe implements PipeTransform {
+
+  public transform(user: User, ...args: unknown[]): string {
```
mikzat, line 9: This pipe is not accepting arguments, so remove `...args: unknown[]` and just leave `user: User`

frontend/rolecall/src/app/common_components/full-name.pipe.ts
```diff
@@ -0,0 +1,19 @@
...
+    if (!user) {
```
mikzat, line 10: The arg type is defined as `user: User`, which means it cannot be null or undefined. Remove this if-check?

mikzat (changes requested):

frontend/rolecall/src/app/cast/cast-editor-v2.component.ts
yhedholm: It was a display bug I discovered while verifying that the pipe worked. It was an easy fix, so I though I could sneak it in.

frontend/rolecall/src/app/common_components/full-name.pipe.ts
yhedholm: Done.

frontend/rolecall/src/app/common_components/full-name.pipe.ts
yhedholm: I want the pipe to be null/undefined sturdy, so I changed the signature to transform(user?: User) { ... }

backend/src/main/java/com/google/rolecall/services/CastServices.java
yhedholm: Sorry! I missed this because it is on the server and I didn't think I had to clean it up before doin the PR.

Removed.

frontend/rolecall/src/app/common_components/full-name.pipe.ts
mikzat: Ok cool, in that case let's use `transform(user: User|undefined|null)`.

There are a couple of issues with using `(user?: User)`
* it is not semantically accurate since the parameter is not actually optional, you always have to pass one into a pipe (even if you pass in undefined or null)
* it does not support potential `null`, since `optional` only means it'll be `User` or `undefined`

yhedholm: Done.

mikzat (approved):

PR #92, opened 2020-11-02
Remove dead code and turn on noUnusedParameters

mikzat:
This removes unused functions, properties, and the CacheValidator service which was never used.

### DIFF
--- a/frontend/rolecall/src/app/api/cast_api.service.ts
+++ b/frontend/rolecall/src/app/api/cast_api.service.ts
@@ -136,10 +136,9 @@ export class CastApi {
               withCredentials: true
             })
         .toPromise()
-        .catch((errorResp) => errorResp)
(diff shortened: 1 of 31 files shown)


yhedholm (approved):
A lot of unnecessary (parameter) parens removed. Really cool! How did you do this? Could I do something like this too?

mikzat:
> A lot of unnecessary (parameter) parens removed. Really cool! How did you do this? Could I do something like this too?

Richard recently turned on a tslint rule to identify single-arg anonymous functions to not allow brackets (which I don't feel too strongly about one way or the other). I found all the instances by running an inspection in my IDE and then fixed them semi-manually. Eventually I believe Richard wants the fixes to happen automatically using tslint CLI but I think we're not quite there yet.

PR #95, opened 2020-11-02
Remove cache service, it's not used.

mikzat:


yhedholm (approved):
Good decision to remove!

PR #99, opened 2020-11-03
Update cloudbuild.unit_tests.yaml

zephyr-l:
Codecov will not be able to access repos under google-intern. Thus we will not upload the coverage information to codecov.


yhedholm (approved):
I am not familiar with this code, but the changes look very straight forward.

mikzat (approved):

PR #101, opened 2020-11-12
fix backend test errors

yhedholm:
I made the backed errors run without errors.

### DIFF
--- a/backend/src/main/java/com/google/rolecall/models/User.java
+++ b/backend/src/main/java/com/google/rolecall/models/User.java
@@ -359,14 +359,14 @@ public class User {
     }
   
     public Builder setNotificationEmail(String notificationEmail) {
-      if(email != null) {
+      if(notificationEmail != null) {
         this.notificationEmail = notificationEmail.toLowerCase();
       }
       return this;
     }
   
     public Builder setPictureFile(String pictureFile) {
-      if(email != null) {
+      if(pictureFile != null) {
         this.pictureFile = pictureFile.toLowerCase();
       }
       return this;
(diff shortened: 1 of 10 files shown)


yhedholm:
Sorry! I meant "I made the backend error tests run without errors".

backend/src/main/java/com/google/rolecall/models/User.java
```diff
@@ -359,14 +359,14 @@ public Builder setEmail(String email) {
...
+      if(notificationEmail != null) {
         this.notificationEmail = notificationEmail.toLowerCase();
       }
       return this;
     }
   
     public Builder setPictureFile(String pictureFile) {
-      if(email != null) {
+      if(pictureFile != null) {
         this.pictureFile = pictureFile.toLowerCase();
```
RichardYang2016, line 370: I believe this should be made lowercase. in URL case is sensitive.

RichardYang2016 (commented):
a potential bug pls check comment. thanks

backend/src/main/java/com/google/rolecall/models/User.java
yhedholm: Done.

RichardYang2016: sorry i had a typo. I was saying this should NOT be made lowercase.

basically you can change to 
  this.pictureFile = pictureFile.
and remove the toLowerCase() call 

Thanks
Richard

RichardYang2016:
sorry i had a typo. I was saying this should NOT be made lowercase.

basically you can change to
this.pictureFile = pictureFile.
and remove the toLowerCase() call

Thanks
Richard

RichardYang2016 (approved):

PR #107, opened 2020-12-30
Turnoff backend unittest

zephyr-l:
When the issue was fixed, please turn the unit tests back on.

Other triggers can be managed in Console - Cloud Build.

Screen Shot 2020-12-29 at 5 03 01 PM

### DIFF
--- a/yaml_cloudbuild/cloudbuild.unit_tests.yaml
+++ b/yaml_cloudbuild/cloudbuild.unit_tests.yaml
@@ -1,16 +1,16 @@
 steps:
 
-# Trigger unit tests for the backend.
-- id: backend_unit_tests
-  name: 'gcr.io/cloud-builders/mvn'
-  args: ['clean', 'test']
-  dir: "backend"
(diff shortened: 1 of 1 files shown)


mikzat (changes requested):
Could you please undo any style changes? It's extremely difficult to identify what lines changed specifically to turn off the unit tests.

zephyr-l:
I am not quite sure if we just want to deploy this into Master/Prod or we would like to "have the fix" AND "merge Master to Prod". I rebased this from Master and now it should work. Any additional changes should have the yaml file included in their branch so that the test can pass.

zephyr-l:
By the way there are front-end testings failing as well. Shall we just remove all the unit-testing checks or we plan to fix the unit tests first?

mikzat (approved):

mikzat:
Thanks Zephyr! 

For now we only agreed to disable the backend tests as the frontend ones were ok last time we checked. We should look into fixing FE tests rather than disabling. But thanks to this PR we now know where to look to disable them if we need to.

PR #108, opened 2021-01-02
fixed two crashes and addressed some customer complaints

yhedholm:


frontend/rolecall/src/app/piece/piece_editor.component.ts
```diff
@@ -231,7 +231,8 @@ export class PieceEditor implements OnInit {
     createPosition: boolean,
     name: string,
   }) {
-    return (createPosition ? '' : 'Ballet ') + name;
+    // You can prefix a ballet by adding text here
+    return (createPosition ? '' : '') + name;
```
mikzat, line 235: This line is the equivalent of:
```
return name;
```

Is that intentional? Can you simplify it?

mikzat (changes requested):

frontend/rolecall/src/app/piece/piece_editor.component.ts
yhedholm: Yes, it is intentional. In this screen you can either edit a Ballet or a Super Ballet. If you are editing a Ballet, the line items are Positions (dance positions that need to be casted). If you are editing a Super Ballet, the line items are Ballets which are different and cannot be casted directly.

In the past, we prefixed the Ballet line items with 'Ballet' to tell the users that they were not dealing with positions. At the last customer call, one customer asked us to remove the 'Ballet' prefix to see how that feels.

I would like to keep the logic so that 1) it is easy to put 'Ballet' back if the customer changes her mind, and 2) keep the logic so that it is easy to prefix base on line item type (Position or Ballet).

mikzat: Got it, in that case let's do something like this:
```
    // TODO: Remove this once the customer confirms it is not needed.
    // return (createPosition ? '' : '') + name;
    return name;
```

yhedholm: Done.

mikzat (approved):

PR #109, opened 2021-01-02
Build standard navigation

yhedholm:
This is a large PR. The customer wasn't happy with the inconsistent navigation we had in the different tabs, especially between the Segments tab and the Casts tab.

I addressed this by building a generic navigation object from the segment navigator (which was better than the cast navigator). In doing this, I had to move a lot of code around, which makes for lots of diffs.

Richard, in my experience with Mikhail, I know that he hates these big PRs (he likes to go really deep on smaller checkins), so I am only asking you to review this one. From my perspective, it is OK if a bug or two slip by. Right now, we only have two users at Alvin Ailey, and they are very forgiving. It is more important that they get to play with the code soon, rather than that we give them perfect code significantly later.


RichardYang2016 (approved):
i did not check line by line but at high level it looks good.

PR #111, opened 2021-01-06
Fix build error

yhedholm:
Mikhail, with the new return statement, there is an unused variable in the call function that breaks the build because the parameter is unused. I reverted to the old return statement.


mikzat (approved):

mikzat:
The unit tests should not be failing for this PR, please make sure you pull from master.

mikzat:
Oh, it looks like something already broke FE unit tests after I've fixed them, unrelated to this PR.

PR #172, opened 2022-10-16
Profile pic from gcs

RichardYang2016:
Resizing is pending. now pic retrieval is from GCS. links are valid for 1 day (signature refreshed when window is refreshed)

Suggest: even for local we should use GCS instead of using local file as storage. everyone should create a separate bucket in GCS and set GCP key file as env by using mac( in terminal)
export GOOGLE_APPLICATION_CREDENTIALS="KEY_PATH"
or if windows (in cmd)
set GOOGLE_APPLICATION_CREDENTIALS=KEY_PATH

Yorgen: personally I like that profile 'dev' has everything local and we don't have to deal with GOOGLE_APPLICATION_CREDENTIALS and can be completely off line. If we implement the suggestion, could we create a new profile, say 'devqa' where we require GOOGLE_APPLICATION_CREDENTIALS and also use the qa server database?


WolfgangEdholm (approved):

backend/src/main/java/com/google/rolecall/services/LocalStorageService.java
```diff
@@ -80,6 +92,18 @@ public Resource loadAsResource(AssetType type, String filename)
...
 
+  @Override
+  public String loadAsUrl(AssetType type, String filename)
+      throws InvalidParameterException, FileNotFoundException {           
+        Blob blob = storage.get(
+          BlobId.of(bucketName, String.format("%s/%s", type.location, filename)));
+        if (blob == null || !blob.exists()) {
+          throw new FileNotFoundException("No file found for " + filename + ".");
+        }
+        return storage.signUrl(blob.toBuilder().setContentType("text/plain").build(), 1, TimeUnit.DAYS, SignUrlOption.withV4Signature()).toString();
```
jaredhirsch-google, line 103: Since this does not function, this should throw an unimplemented exception.

jaredhirsch-google (approved):
Just one small tweak to the backend side of things.

backend/src/main/java/com/google/rolecall/services/LocalStorageService.java
RichardYang2016: done

